PR #1, opened 2020-05-27
repo setup including auth changes

Sravanksk:


apps/mappings/migrations/0001_initial.py
```diff
@@ -0,0 +1,77 @@
+# Generated by Django 3.0.3 on 2020-06-03 04:14
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
```
Shwetabhk, line 7: All of this is not needed with generic mapping infra

apps/mappings/models.py
```diff
@@ -0,0 +1,71 @@
+"""
+Mapping Models
+"""
+from django.db import models
+
+from apps.workspaces.models import Workspace
+
+
+class EmployeeMapping(models.Model):
```
Shwetabhk, line 9: We dont need mapping tables

PR #6, opened 2020-07-14
return limited data from netsuite api authentication 

Sravanksk:


apps/workspaces/views.py
```diff
@@ -174,7 +174,13 @@ def post(self, request, **kwargs):
                 netsuite_credentials.save()
 
             return Response(
-                data=NetSuiteCredentialSerializer(netsuite_credentials).data,
+                data={
```
Shwetabhk, line 177: mentioning each field manually, you can edit the serializer to choose the fields you want to serialize

apps/workspaces/views.py
```diff
@@ -203,11 +209,17 @@ def get(self, request, **kwargs):
...
             workspace = Workspace.objects.get(pk=kwargs['workspace_id'])
-            fyle_credentials = NetSuiteCredentials.objects.get(workspace=workspace)
+            netsuite_credentials = NetSuiteCredentials.objects.get(workspace=workspace)
 
-            if fyle_credentials:
+            if netsuite_credentials:
                 return Response(
-                    data=NetSuiteCredentialSerializer(fyle_credentials).data,
+                    data={
+                        'id': netsuite_credentials.id,
```
Shwetabhk, line 217: same

PR #9, opened 2020-08-19
migrate netsuite to fyle-rest-auth==0.5.0

Sravanksk:


fyle_netsuite_api/settings.py
```diff
@@ -125,6 +125,13 @@
     },
 }
 
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
+        'LOCATION': 'cache_table',
+    }
```
Sravanksk, line 132: `cache_table` will be the name of the table in the DB

Shwetabhk: rename to `auth_cache`

Shwetabhk (approved):

PR #14, opened 2020-09-21
support for default CCC vendor and exporting CCC expenses as bills

Sravanksk:


apps/netsuite/models.py
```diff
@@ -104,20 +105,26 @@ def create_bill(expense_group: ExpenseGroup):
...
         general_mappings = GeneralMapping.objects.get(workspace_id=expense_group.workspace_id)
         subsidiary_mappings = SubsidiaryMapping.objects.get(workspace_id=expense_group.workspace_id)
 
-        bill_object, _ = Bill.objects.update_or_create(
-            expense_group=expense_group,
-            defaults={
-                'subsidiary_id': subsidiary_mappings.internal_id,
-                'accounts_payable_id': general_mappings.accounts_payable_id,
-                'vendor_id': Mapping.objects.get(
+        if general_settings.corporate_credit_card_expenses_object == 'BILL':
```
Shwetabhk, line 112: If someone is uploading both reimbursable and non-reimbursable expenses as bills
the else case will never be executed right?

Shwetabhk (changes requested):

Shwetabhk (approved):

PR #15, opened 2020-09-28
support fyle expense custom-fields and advanced import settings

Sravanksk:


apps/mappings/models.py
```diff
@@ -27,8 +27,6 @@ class GeneralMapping(models.Model):
     General Mapping
     """
     id = models.AutoField(primary_key=True)
-    location_name = models.CharField(max_length=255, help_text='NetSuite Location name', null=True)
```
Shwetabhk, line 30: No need to remove this

Shwetabhk (approved):

Shwetabhk (approved):

PR #25, opened 2020-11-26
Auto import and Auto map NetSuite Projects into Fyle

ashwin1111:


apps/mappings/tasks.py
```diff
@@ -0,0 +1,146 @@
...
+    :param projects: NetSuite Projects
+    :param workspace_id: integer id of workspace
+    :return: Fyle Projects Payload
+    """
+    payload = []
+    existing_project_names = ExpenseAttribute.objects.filter(
+        attribute_type='PROJECT', workspace_id=workspace_id).values_list('value', flat=True)
+
+    for project in projects:
+        if project['value'] not in existing_project_names and project['parent_name'] not in existing_project_names:
```
Shwetabhk, line 30: What if there is a new project under a parent name?

Shwetabhk:
```
Few things missing -
* Billable expenses are not being marked as billable in NetSuite
* Check support in NetSuite SDK
* Check this commit https://github.com/fylein/fyle-qbo-api/commit/85b8821aa20b1d394ae5d593e772f16b1264b24e

File: apps/mappings/tasks.py (hunk 1 above)
Comment (Shwetabhk) [reply apps/mappings/tasks.py:30]: eg you have a project A
under this there are 2 Subprojects X and Y

If A / X is already saved
you'll end up skipping A/Y

Comment (Shwetabhk) [reply apps/mappings/tasks.py:30]: if 'Parent / Value' exists then dont create else create

File: apps/mappings/tasks.py [hunk 2]
```diff
@@ -0,0 +1,137 @@
...
+    ns_connection = NetSuiteConnector(
+        netsuite_credentials=ns_credentials,
+        workspace_id=workspace_id
+    )
+
+    fyle_connection.sync_projects()
+
+    ns_connection.sync_projects()
+
+    # ns_connection.sync_customers()
```
ashwin1111, line 66: ignore for now, takes >5 mins to test

apps/mappings/tasks.py
```diff
@@ -0,0 +1,137 @@
...
+    # TODO: remove comment
+
+    ns_attributes = DestinationAttribute.objects.filter(attribute_type='PROJECT', workspace_id=workspace_id).all()
+
+    fyle_payload: List[Dict] = create_fyle_projects_payload(ns_attributes, workspace_id)
+    if fyle_payload:
+        # TODO: rm print
+        print('fyle_payload', fyle_payload)
```
ashwin1111, line 74: will remove after testing

apps/netsuite/models.py
```diff
@@ -330,7 +361,9 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
                         'TRANSACTION_LINE', 'ALL'] else None,
                     'class_id': class_id,
                     'department_id': department_id,
+                    'customer_id': customer_id if lineitem.billable else None,
```
Shwetabhk, line 364: hey this is wrong no? there might be a customer id even if billable is false?

apps/netsuite/models.py
```diff
@@ -500,9 +540,10 @@ def create_expense_report_lineitems(expense_group: ExpenseGroup):
                 expense_id=lineitem.id,
                 defaults={
                     'amount': lineitem.amount,
+                    'billable': lineitem.billable if customer_id else None,
                     'category': account.destination.destination_id,
                     'class_id': class_id if class_id else None,
-                    'customer_id': None,
+                    'customer_id': customer_id if lineitem.billable else None,
```
Shwetabhk, line 546: hey this is wrong no? there might be a customer id even if billable is false?

apps/fyle/utils.py
```diff
@@ -211,15 +211,26 @@ def sync_cost_centers(self, active_only: bool):
...
         """
         Get projects from fyle
         """
-        projects = self.connection.Projects.get(active_only=active_only)['data']
+        all_projects = []
+        limit = 1000
+        offset = 0
+
+        while True:
+            projects = self.connection.Projects.get(limit=str(limit), offset=str(offset))['data']
```
Shwetabhk, line 223: hey does get of projects have pagination?

apps/mappings/tasks.py
```diff
@@ -0,0 +1,134 @@
...
+                destination_type='PROJECT',
+                source_value=project.value,
+                destination_value=project.value,
+                workspace_id=workspace_id
+            )
+            project_mappings.append(mapping)
+
+        return project_mappings
+    except WrongParamsError as exception:
+        logger.exception(
```
Shwetabhk, line 100: use logger.error

apps/mappings/tasks.py
```diff
@@ -0,0 +1,134 @@
...
+            'Error while creating projects workspace_id - %s in Fyle %s %s',
+            workspace_id, exception.message, {'error': exception.response}
+        )
+    except Exception:
+        error = traceback.format_exc()
+        error = {
+            'error': error
+        }
+        logger.exception(
```
Shwetabhk, line 109: use logger.error

apps/mappings/tasks.py
```diff
@@ -0,0 +1,134 @@
...
+
+def schedule_projects_creation(import_projects, workspace_id):
+    if import_projects:
+        start_datetime = datetime.now()
+        schedule, _ = Schedule.objects.update_or_create(
+            func='apps.mappings.tasks.auto_create_project_mappings',
+            args='{}'.format(workspace_id),
+            defaults={
+                'schedule_type': Schedule.MINUTES,
+                'minutes': 6 * 60,
```
Shwetabhk, line 123: make it once every 24 hours

apps/fyle/utils.py
ashwin1111: yes, it supports passing of offset limit. But we don't get any count of number of projects in response.
Sample response:
```{'offset': 0, 'limit': 1000, 'url': '/api/tpa/v1/projects', 'resource_type': 'Project', 'data':[{},{},..]}```

Comment (ashwin1111) [reply apps/fyle/utils.py:223]: https://app.swaggerhub.com/apis-docs/F588/Fyle_TPA/0.1#/Projects/get_api_tpa_v1_projects here we have count.

File: apps/netsuite/models.py (hunk 4 above)
Comment (ashwin1111) [reply apps/netsuite/models.py:364]: If billable is false and we are sending some customer in payload, then it is automatically converted to a Billable expense :)

File: apps/netsuite/models.py (hunk 5 above)
Comment (ashwin1111) [reply apps/netsuite/models.py:546]: If billable is false and we are sending some customer in payload, then it is automatically converted to a Billable expense :)

File: apps/fyle/utils.py (hunk 6 above)
Comment (Shwetabhk) [reply apps/fyle/utils.py:223]: Doesn't it return all the projects at once?

File: apps/netsuite/models.py (hunk 5 above)
Comment (Shwetabhk) [reply apps/netsuite/models.py:546]: Then send billable as false no?

File: apps/netsuite/models.py (hunk 4 above)
Comment (Shwetabhk) [reply apps/netsuite/models.py:364]: then send billable as false no?

File: apps/fyle/utils.py (hunk 6 above)
Comment (ashwin1111) [reply apps/fyle/utils.py:223]: No, it returns max of 1000 projects in one call

Comment (Shwetabhk) [reply apps/fyle/utils.py:223]: cool, then we need to fix this everywhere :fml:

Comment (Shwetabhk) [reply apps/fyle/utils.py:223]: or we can do a hacky solution in Fyle SDK for now

Comment (ashwin1111) [reply apps/fyle/utils.py:223]: yes, anything works

File: apps/netsuite/models.py [hunk 10]
```diff
@@ -352,6 +352,13 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
             if general_settings.import_projects:
                 customer_id = get_customer_id_or_none(expense_group, lineitem)
 
+            billable = lineitem.billable
+            if billable is None and customer_id:
+                billable = False
+
+            if customer_id is None and billable:
```
Shwetabhk, line 359: What if billable is True here?

apps/netsuite/models.py
```diff
@@ -321,6 +348,17 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
...
 
+            customer_id = None
+            if general_settings.import_projects:
+                customer_id = get_customer_id_or_none(expense_group, lineitem)
+
+            billable = lineitem.billable
+            if billable is None and customer_id:
+                billable = False
+
+            if customer_id is None and billable:
```
Shwetabhk, line 359: this condition checks
`if customer_id == null and billable == False {}`
what if `customer_id = null` and `billable = True`
the condition fails, which means the value of billable is still `True` and not changed to `null`

we can rather replace block 355-360 with
```python
if customer_id:
    if not billable:
        billable = False
else:
    billable = None
```

apps/netsuite/models.py
```diff
@@ -495,14 +542,22 @@ def create_expense_report_lineitems(expense_group: ExpenseGroup):
 
             custom_segments = get_custom_segments(expense_group, lineitem)
 
+            billable = lineitem.billable
```
Shwetabhk, line 545: same

apps/netsuite/models.py
ashwin1111: ```what if customer_id = null and billable = True```

Intentionally assigning billable as None

Else, export would fail if customer id is null and billable is true.

ashwin1111: okay ignore above comment, both of them are almost same. Will test and update PR.

apps/netsuite/models.py
ashwin1111: done

Shwetabhk (approved):

Sravanksk (approved):

PR #28, opened 2020-12-08
ignore accounts_payable account when creating vendor payments

Sravanksk:


apps/netsuite/utils.py
```diff
@@ -941,7 +941,7 @@ def __construct_vendor_payment(self, vendor_payment: VendorPayment,
             'balance': None,
             'apAcct': {
```
Shwetabhk, line 942: cant this be changed to `apAcct: None`

scripts/005-update-payment-sync-status.sql
```diff
@@ -0,0 +1,18 @@
+-- Script to set payment_synced as True for existing bills, expense_reports and journal_entries
```
Shwetabhk, line 1: Since this is not on prod yet, can you add it to the existing script? 004

Shwetabhk (approved):

PR #31, opened 2020-12-15
mark expense as CCC in expensereports if expense fund_source is CCC

Sravanksk:


apps/netsuite/utils.py
```diff
@@ -542,7 +546,7 @@ def __construct_expense_report_lineitems(
                     'externalId': None,
                     'type': 'account'
                 },
-                'corporateCreditCard': None,
+                'corporateCreditCard': is_ccc,
```
Shwetabhk, line 549: Change this to
`True if expense.fund_source === 'CCC' else False`

apps/netsuite/utils.py
```diff
@@ -513,6 +513,10 @@ def __construct_expense_report_lineitems(
         for line in expense_report_lineitems:
             expense = Expense.objects.get(pk=line.expense_id)
 
+            is_ccc = None
```
Shwetabhk, line 516: extra variable is not needed

PR #47, opened 2021-02-08
Auto map employees Fyle <> Netsuite

ashwin1111:


apps/fyle/utils.py
```diff
@@ -156,11 +156,16 @@ def sync_employees(self):
         employee_attributes = []
 
         for employee in employees:
+            detail = {
+                'employee_code': employee['employee_code'],
```
Shwetabhk, line 160: can you also add `department` and `location` for future use?

apps/workspaces/models.py
```diff
@@ -85,6 +85,8 @@ class WorkspaceGeneralSettings(models.Model):
                                                          help_text='Auto Sync Payments from Fyle to Netsuite')
     sync_netsuite_to_fyle_payments = models.BooleanField(default=False,
                                                          help_text='Auto Sync Payments from NetSuite to Fyle')
+    auto_map_employees = models.CharField(max_length=50, help_text='Auto Map Employees type from \
```
ashwin1111, line 88: will fix this

apps/mappings/tasks.py
```diff
@@ -132,3 +133,61 @@ def schedule_projects_creation(import_projects, workspace_id):
 
         if schedule:
             schedule.delete()
+
+def async_auto_map_employees(auto_map_option: str, general_settings: WorkspaceGeneralSettings, workspace_id: str):
```
Shwetabhk, line 137: can we rename the variable to `employee_mapping_preference`

apps/mappings/tasks.py
```diff
@@ -132,3 +133,61 @@ def schedule_projects_creation(import_projects, workspace_id):
...
+    mapping_setting = MappingSetting.objects.filter(
+        ~Q(destination_field='CREDIT_CARD_ACCOUNT'),
+        source_field='EMPLOYEE', workspace_id=workspace_id
```
Shwetabhk, line 140: Vendor Case is not handled

apps/mappings/tasks.py
```diff
@@ -132,3 +133,61 @@ def schedule_projects_creation(import_projects, workspace_id):
...
+                }
+
+        elif auto_map_option == 'NAME':
+            filters = {
+                'detail__contains': {'full_name': employee.value}
+            }
+
+        elif auto_map_option == 'EMPLOYEE_CODE':
+            filters = {
+                'detail__contains': {'employee_code': employee.value}
```
Shwetabhk, line 166: cant we filter directly with JSON Key? instead of using contains

apps/mappings/tasks.py
```diff
@@ -132,3 +133,61 @@ def schedule_projects_creation(import_projects, workspace_id):
...
+    for employee in employee_attributes:
+        if auto_map_option == 'EMAIL':
+            if employee.detail and employee.detail['email']:
+                filters = {
+                    'value': employee.detail['email']
+                }
+
+        elif auto_map_option == 'NAME':
+            filters = {
+                'detail__contains': {'full_name': employee.value}
```
Shwetabhk, line 161: cant we filter directly with JSON Key? instead of using contains

apps/mappings/tasks.py
ashwin1111: source_field will always be EMPLOYEE (Fyle)

apps/mappings/tasks.py
ashwin1111: yes, will check

apps/mappings/tasks.py
```diff
@@ -132,3 +133,61 @@ def schedule_projects_creation(import_projects, workspace_id):
...
+
+        elif auto_map_option == 'EMPLOYEE_CODE':
+            filters = {
+                'detail__contains': {'employee_code': employee.value}
+            }
+
+        if filters:
+            source_attributes = ExpenseAttribute.objects.filter(workspace_id=workspace_id, **filters).all()
+
+        for source in source_attributes:
```
Shwetabhk, line 172: are we mappings multiple source attributes to the same destination attributes?

ashwin1111: incase if `source_attributes` returns more than 1.
for test orgs, there can be email repetitions / name repetitions. So, to avoid confusion added this, on prod, `source_attributes` will be of length 1 in almost all cases.

apps/mappings/tasks.py
Shwetabhk: Ahhh ok, yeah

apps/fyle/utils.py
```diff
@@ -160,7 +160,15 @@ def sync_employees(self):
...
                 'value': employee['employee_email'],
-                'source_id': employee['id']
+                'source_id': employee['id'],
+                'detail': {
+                    'employee_code': employee['employee_code'],
+                    'full_name': employee['full_name'],
+                    'location': employee['location'],
+                    'department': employee['department'],
+                    'department_id': employee['department_id'],
+                    'department_code': employee['department_code']
```
ashwin1111, line 170: TODO: Shwetabh to comment on unnecessary keys to remove.

Shwetabhk: This looks fine

Shwetabhk (approved):

Shwetabhk (approved):

Sravanksk (approved):

PR #55, opened 2021-02-17
adding tests for fyle and workspace apps

Sravanksk:
Test Cases covered: -

Models - 
`Added tests for all the Models in every App`

Views - 
`Added tests for all GET and POST calls in all the Views except schedule API's`

TO-DO 
`Write tests for all schedule related API's endpoints`

cc @Shwetabhk


apps/fyle/tests/test_views.py
```diff
@@ -0,0 +1,119 @@
...
+        client_secret = settings.FYLE_CLIENT_SECRET
+        base_url = settings.FYLE_BASE_URL
+        refresh_token = settings.FYLE_REFRESH_TOKEN
+
+        self.connection = FyleSDK(
+            base_url=base_url,
+            client_id=client_id,
+            client_secret=client_secret,
+            refresh_token=refresh_token,
+            jobs_url=settings.FYLE_JOBS_URL
```
Shwetabhk, line 26: why do we need Jobs URL?

apps/fyle/tests/test_views.py
```diff
@@ -0,0 +1,119 @@
...
+            refresh_token=refresh_token,
+            user=self.user
+        )
+        self.auth_token.save()
+
+        self.client = APIClient()
+        self.api_authentication()
+
+    def api_authentication(self):
+        self.client.credentials(HTTP_AUTHORIZATION='Bearer ' + self.access_token)
```
Shwetabhk, line 46: use string formatting instead of concatenation

apps/fyle/tests/test_views.py
```diff
@@ -0,0 +1,119 @@
...
+
+class FyleTests(APITestCase):
+
+    def setUp(self):
+        client_id = settings.FYLE_CLIENT_ID
+        client_secret = settings.FYLE_CLIENT_SECRET
+        base_url = settings.FYLE_BASE_URL
+        refresh_token = settings.FYLE_REFRESH_TOKEN
+
+        self.connection = FyleSDK(
```
Shwetabhk, line 21: See if it is possible to create all the authentication stuff as a function in a new file `fyle_netsuite_api/test_utils.py`
and then we can import and use from there in all the apps

apps/fyle/tests/test_views.py
```diff
@@ -0,0 +1,119 @@
...
+            id=1,
+            refresh_token=refresh_token,
+            user=self.user
+        )
+        self.auth_token.save()
+
+        self.client = APIClient()
+        self.api_authentication()
+
+    def api_authentication(self):
```
Shwetabhk, line 45: Let's move this to `fyle_netsuite_api/test_utils.py`

fyle_netsuite_api/settings.py
```diff
@@ -242,6 +242,7 @@
 FYLE_CLIENT_SECRET = os.environ.get('FYLE_CLIENT_SECRET')
 FYLE_BASE_URL = os.environ.get('FYLE_BASE_URL')
 FYLE_JOBS_URL = os.environ.get('FYLE_JOBS_URL')
+FYLE_REFRESH_TOKEN = os.environ.get('FYLE_REFRESH_TOKEN')
```
Shwetabhk, line 245: Don't import Refresh token in settings since it is only used for testing

apps/fyle/tests/test_views.py
```diff
@@ -0,0 +1,119 @@
...
+from rest_framework.test import APITestCase, APIClient
+
+from apps.workspaces.models import Workspace
+from fyle_netsuite_api import settings
+
+
+class FyleTests(APITestCase):
+
+    def setUp(self):
+        client_id = settings.FYLE_CLIENT_ID
```
Shwetabhk, line 16: Dont use the same creds as the app for testing
do something like 
`client_id = os.environ.get('FYLE_TEST_CLIENT_ID')`

apps/fyle/tests/test_views.py
Sravanksk: 👍

fyle_netsuite_api/test_settings.py
```diff
@@ -0,0 +1,250 @@
+"""
+Django settings for fyle_netsuite_api project.
```
Shwetabhk, line 2: Why do we need an extra settings file for this?

PR #56, opened 2021-02-18
Auto Sync of Categories in Netsuite

Shwetabhk:


apps/netsuite/utils.py
```diff
@@ -110,21 +110,28 @@ def sync_expense_categories(self):
...
+                'account_internal_id': category['expenseAcct']['internalId']
+            }
+
             category_attributes.append(
                 {
                     'attribute_type': 'EXPENSE_CATEGORY',
                     'display_name': 'Expense Category',
-                    'value': 'Expense Category - {}'.format(category['name']),
-                    'destination_id': category['internalId']
+                    'value': category['name'],
```
ashwin1111, line 122: spl character fix if needed

apps/netsuite/utils.py
```diff
@@ -110,21 +110,28 @@ def sync_expense_categories(self):
...
                 }
             )
 
             category_attributes.append(
                 {
                     'attribute_type': 'CCC_EXPENSE_CATEGORY',
                     'display_name': 'Credit Card Expense Category',
-                    'value': 'Expense Category - {}'.format(category['name']),
-                    'destination_id': category['internalId']
+                    'value': category['name'],
```
ashwin1111, line 132: same

apps/mappings/tasks.py
```diff
@@ -30,6 +30,209 @@ def remove_duplicates(ns_attributes: List[DestinationAttribute]):
     return unique_attributes
 
 
+def create_fyle_categories_payload(categories: List[DestinationAttribute], workspace_id: int):
+    """
+    Create Fyle Categories Payload from QBO Customer / Categories
+    :param workspace_id: Workspace integer id
+    :param categories: QBO Categories
```
ashwin1111, line 37: NetSuite

apps/mappings/tasks.py
```diff
@@ -30,6 +30,209 @@ def remove_duplicates(ns_attributes: List[DestinationAttribute]):
     return unique_attributes
 
 
+def create_fyle_categories_payload(categories: List[DestinationAttribute], workspace_id: int):
+    """
+    Create Fyle Categories Payload from QBO Customer / Categories
```
ashwin1111, line 35: NetSuite

apps/mappings/tasks.py
```diff
@@ -30,6 +30,209 @@ def remove_duplicates(ns_attributes: List[DestinationAttribute]):
...
+                    'destination_value': category.value,
+                    'destiantion_type': reimbursable_destination_type
+                }
+                logger.error(
+                    'Error while creating categories auto mapping workspace_id - %s %s',
+                    workspace_id, {'payload': detail}
+                )
+                raise ExpenseAttribute.DoesNotExist
+
+            create_credit_card_category_mappings(
```
Shwetabhk, line 194: This is an extra step as compared to QBO, please check this out fully

apps/mappings/tasks.py
```diff
@@ -30,6 +30,209 @@ def remove_duplicates(ns_attributes: List[DestinationAttribute]):
...
+
+    netsuite_attributes = remove_duplicates(netsuite_attributes)
+
+    fyle_payload: List[Dict] = create_fyle_categories_payload(netsuite_attributes, workspace_id)
+
+    if fyle_payload:
+        fyle_connection.connection.Categories.post(fyle_payload)
+
+        if reimbursable_expenses_object == 'EXPENSE REPORT':
+            netsuite_connection.sync_expense_categories()
```
ashwin1111, line 87: why do we need to sync netsuite data again?
Ideally we need to sync fyle categories after POST, to retrieve all fyle attributes to expense_attribute table right?

apps/mappings/tasks.py
```diff
@@ -30,6 +30,209 @@ def remove_duplicates(ns_attributes: List[DestinationAttribute]):
...
+                Mapping.create_or_update_mapping(
+                    source_type='CATEGORY',
+                    destination_type='CCC_EXPENSE_CATEGORY',
+                    source_value=mapping.source.value,
+                    destination_value=mapping.destination.value,
+                    destination_id=mapping.destination.destination_id,
+                    workspace_id=workspace_id
+                )
+
+        elif corporate_credit_card_expenses_object in ('BILL', 'JOURNAL ENTRY'):
```
ashwin1111, line 114: `else` would do the job imho

Shwetabhk: else would also include `None`

apps/mappings/tasks.py
Shwetabhk: Hmm it is this way in QBO too, I wonder why
will change it

apps/mappings/tasks.py
Shwetabhk: noted

apps/mappings/tasks.py
Shwetabhk: noted

ashwin1111 (approved):

Sravanksk (approved):

PR #63, opened 2021-03-01
Auto create employee / vendor

Sravanksk:


apps/netsuite/utils.py
```diff
@@ -348,6 +416,68 @@ def sync_employees(self):
...
+                auto_map_employee_preference == 'EMPLOYEE_CODE' and employee.detail['employee_code']
+        ) else employee.detail['full_name']
+
+        employee = {
+            'entityId': employee_entity_id,
+            'firstName': employee.detail['full_name'].split(' ')[0],
+            'lastName': employee.detail['full_name'].split(' ')[-1]
+            if len(employee.detail['full_name'].split(' ')) > 1 else '',
+            'inheritIPRules': True,
+            'payFrequency': '_monthly',
```
ashwin1111, line 445: what does this mean

apps/netsuite/utils.py
```diff
@@ -348,6 +416,68 @@ def sync_employees(self):
...
+            'subsidiary': {
+                'name': None,
+                'internalId': subsidiary_mapping.internal_id,
+                'externalId': None,
+                'type': None
+            },
+            'workCalendar': {
+                "name": None,
+                "internalId": 1,
```
ashwin1111, line 454: why is it hard coded as 1?

ashwin1111: also, are both `'1' or 1` acceptable by netsuite?

apps/netsuite/utils.py
```diff
@@ -348,6 +416,68 @@ def sync_employees(self):
...
+            'defaultExpenseReportCurrency': {
+                "internalId": currency.destination_id if currency else '1',
+                "externalId": None,
+                "type": "currency"
+            },
+            'externalId': employee.value
+        }
+        created_employee = self.connection.employees.post(employee)
+
+        employee_detail = self.connection.employees.get(created_employee['internalId'])
```
ashwin1111, line 467: why do we need this call?
We already have `entityId, internalId, email` right even without this call right?

requirements.txt
```diff
@@ -27,4 +27,4 @@ six==1.13.0
 sqlparse==0.3.0
 Unidecode==1.1.2
 urllib3==1.25.7
-wcwidth==0.1.8
```
ashwin1111, line 30: don't we need to bump up netsuite sdk version?

apps/netsuite/utils.py
```diff
@@ -312,6 +312,74 @@ def sync_vendors(self):
             vendor_attributes, self.workspace_id)
         return vendor_attributes
 
+    def post_vendor(self, vendor: ExpenseAttribute, auto_map_employee_preference: str, expense_group: ExpenseGroup):
```
ashwin1111, line 315: most of the parts is same in `post_employees` and `post_vendors`, see if you can write a separate function for common parts

apps/netsuite/tasks.py
```diff
@@ -70,19 +70,78 @@ def load_attachments(netsuite_connection: NetSuiteConnector, expense_id: str, ex
...
+
+        except NetSuiteRequestError as exception:
+            logger.exception({'error': exception})
+
+        except Exception:
+            error = traceback.format_exc()
+            detail = {
+                'error': error
+            }
+            logger.exception('Something unexpected happened workspace_id: %s %s', expense_group.workspace_id, detail)
```
ashwin1111, line 123: `Something unexpected happened during auto create employee / vendor workspace . . . ` would be easy for us to debug

ashwin1111: does is needs to be `logger.error` or `logger.exception`?

apps/netsuite/utils.py
Sravanksk: https://www.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2019_1/schema/enum/employeepayfrequency.html?mode=package

please to check this

apps/netsuite/utils.py
Sravanksk: `workCalendar` is a required field - setting it to 1 by default, will check for any other possibilities here

apps/netsuite/tasks.py
Sravanksk: this should be fine

apps/netsuite/utils.py
Shwetabhk: set to None

apps/netsuite/utils.py
Shwetabhk: set `internalId` to None

ashwin1111 (approved):

PR #65, opened 2021-03-05
Remove start_now date in API & add auto_mapped true for auto project mappings

KirtiGautam:
- Auto_mapped = True for auto project mappings
- Removed start_now in API and hardcoded it in functions


apps/workspaces/tasks.py
```diff
@@ -10,15 +10,14 @@
 from apps.workspaces.models import WorkspaceSchedule, WorkspaceGeneralSettings
 
 
-def schedule_sync(workspace_id: int, schedule_enabled: bool, hours: int, next_run: str):
```
ashwin1111, line 13: This is a critical area, make sure you test schedule well.
Create new / Switch off / Update etc

ashwin1111 (approved):

Shwetabhk (approved):

PR #70, opened 2021-03-11
BugFix: Map Employee/Vendor before creating

Sravanksk:


apps/netsuite/tasks.py
```diff
@@ -93,12 +93,37 @@ def create_or_update_employee_mapping(expense_group: ExpenseGroup, netsuite_conn
...
+
+                if employee is None:
+                    created_entity: DestinationAttribute = netsuite_connection.post_employee(
+                        source_employee, auto_map_employees_preference, expense_group)
+
             else:
-                created_entity: DestinationAttribute = netsuite_connection.post_vendor(
-                    source_employee, auto_map_employees_preference, expense_group)
+                vendor = DestinationAttribute.objects.filter(
+                    attribute_type='VENDOR',
```
Shwetabhk, line 119: why not add this to filters itself, since it is condtional?

apps/netsuite/tasks.py
```diff
@@ -93,12 +93,37 @@ def create_or_update_employee_mapping(expense_group: ExpenseGroup, netsuite_conn
...
+            elif auto_map_employees_preference == 'NAME':
+                filters = {
+                    'value__iexact': source_employee.detail['full_name']
+                }
+
             if employee_mapping_setting == 'EMPLOYEE':
-                created_entity: DestinationAttribute = netsuite_connection.post_employee(
-                    source_employee, auto_map_employees_preference, expense_group)
+                employee = DestinationAttribute.objects.filter(
+                    attribute_type='EMPLOYEE',
```
ashwin1111, line 108: sir, attribute type should be employee / vendor

Shwetabhk, line 108: why not add this to filters itself, since it is conditional?

ashwin1111: ignore

Shwetabhk: That way you'll have to write the query code only once

Shwetabhk (approved):

PR #77, opened 2021-03-22
Migrate dashboard sync calls to backend

ashwin1111:


apps/fyle/views.py
```diff
@@ -364,3 +365,69 @@ def post(self, request, *args, **kwargs):
...
+    Sync Fyle Dimensions view
+    """
+    def post(self, request, *args, **kwargs):
+        """
+        Sync data from Fyle
+        """
+        try:
+            workspace = Workspace.objects.get(id=kwargs['workspace_id'])
+            if workspace.source_synced_at:
+                time_interval = datetime.now(timezone.utc) - workspace.source_synced_at
```
ashwin1111, line 380: datetime.now() -> `2021-03-22 05:23:31.043255`
datetime.now(timezone.utc) -> `2021-03-22 05:23:31.054558+00:00`
from db -> `2021-03-22 05:12:35.144435+00:00`

We cannot subtract datetime.now() and db, so added `timezone.utc`

apps/fyle/views.py
```diff
@@ -364,3 +365,69 @@ def post(self, request, *args, **kwargs):
...
+
+            if workspace.source_synced_at is None or time_interval.days > 0:
+                fyle_credentials = FyleCredential.objects.get(workspace_id=kwargs['workspace_id'])
+                fyle_connector = FyleConnector(fyle_credentials.refresh_token, kwargs['workspace_id'])
+
+                fyle_connector.sync_dimensions()
+
+                workspace.source_synced_at = datetime.now()
+                workspace.save(update_fields=['source_synced_at'])
+                workspace = Workspace.objects.get(id=kwargs['workspace_id'])
```
ashwin1111, line 390: Will remove this off once frontend logic is changed. Reason for adding this:
```
* Fyle and NetSuite sync calls are executed parallely
* NetSuite call takes more time compared to Fyle, the old `source_synced_at` time would be returned from this api call, which shows incorrect last synced at time, hence adding workspace get once all these sync are completed

apps/fyle/views.py
Shwetabhk: cool

Shwetabhk (approved):

Shwetabhk (approved):

PR #82, opened 2021-03-30
Add org_id in Expense Model

KirtiGautam:


ashwin1111 (approved):

KirtiGautam:
Do we keep org_id `NULL`  for expense before this update? @Shwetabhk @ashwin1111 @Sravanksk

Shwetabhk (approved):

PR #88, opened 2021-04-06
Post and Auto Map Projects in Batches

ashwin1111:


apps/mappings/tasks.py
```diff
@@ -262,72 +262,68 @@ def create_fyle_projects_payload(projects: List[DestinationAttribute], workspace
...
+        ns_connection = NetSuiteConnector(
+            netsuite_credentials=ns_credentials,
+            workspace_id=workspace_id
+        )
+
+        fyle_connection.sync_projects()
+        ns_connection.sync_projects()
+        ns_connection.sync_customers()
+
+        post_projects_in_chunks(fyle_connection, workspace_id)
```
Shwetabhk, line 319:
```suggestion
        post_projects_in_batches(fyle_connection, workspace_id)
```

apps/mappings/tasks.py
```diff
@@ -262,72 +262,68 @@ def create_fyle_projects_payload(projects: List[DestinationAttribute], workspace
...
+            destination_id=project.destination_id,
+            workspace_id=workspace_id
+        )
+        mapping.source.auto_mapped = True
+        mapping.source.save()
 
-    ns_connection.sync_customers()
 
-    ns_attributes = DestinationAttribute.objects.filter(attribute_type='PROJECT', workspace_id=workspace_id).all()
+def post_projects_in_chunks(fyle_connection: FyleConnector, workspace_id: int):
```
Shwetabhk, line 279:
```suggestion
def post_projects_in_batches(fyle_connection: FyleConnector, workspace_id: int):
```

apps/mappings/tasks.py
```diff
@@ -262,72 +262,68 @@ def create_fyle_projects_payload(projects: List[DestinationAttribute], workspace
...
+    ns_attributes_count = DestinationAttribute.objects.filter(attribute_type='PROJECT', workspace_id=workspace_id).count()
+    page_size = 200
 
-    ns_attributes = remove_duplicates(ns_attributes)
+    for offset in range(0, ns_attributes_count, page_size):
+        paginated_ns_attributes = DestinationAttribute.objects.filter(
+            attribute_type='PROJECT', workspace_id=workspace_id).order_by('value', 'id')[offset:offset+page_size]
```
Shwetabhk, line 285: Are we sure that we're not querying all the projects here? and then using array slicing?
Lets only query 200 projects at a time no

Shwetabhk: or is this the way to do it?

apps/mappings/tasks.py
```diff
@@ -262,72 +262,68 @@ def create_fyle_projects_payload(projects: List[DestinationAttribute], workspace
...
-
-    ns_connection = NetSuiteConnector(
-        netsuite_credentials=ns_credentials,
-        workspace_id=workspace_id
-    )
-
-    fyle_connection.sync_projects()
-
-    ns_connection.sync_projects()
+def auto_map_projects(ns_attributes: list, workspace_id: int):
```
Shwetabhk, line 265: Can we skip the already mapped projects here? Like we do for Employees
Reason for doing this -
If I manually change the mapping for one of the projects, it will overwrite that.
We should also do this for categories

ashwin1111:
taken care here https://github.com/fylein/fyle-netsuite-api/pull/90

PR #93, opened 2021-04-16
Cache workspace permissions

ashwin1111:


apps/workspaces/permissions.py
```diff
@@ -1,4 +1,6 @@
+from time import time
```
ashwin1111, line 1: TODO: remove loggers before merging

PR #99, opened 2021-04-21
Use generic auto map function

ashwin1111:


apps/mappings/tasks.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import traceback
 from datetime import datetime
+from time import time
```
ashwin1111, line 4: TODO: remove before merging

apps/mappings/tasks.py
```diff
@@ -385,36 +335,19 @@ def async_auto_map_employees(workspace_id: int):
...
-            'destination_value': employee.value,
-            'destination_id': employee.destination_id,
-            'workspace_id': workspace_id
-        }
-
-        auto_create_employee_mappings(source_attributes, mapping_attributes)
+    start = time()
+    Mapping.auto_map_employees('EMPLOYEE', destination_type, employee_mapping_preference, workspace_id)
+    end = time()
+    print('Total time taken is', end - start)
```
Sravanksk, line 346: rm

Sravanksk (approved):

PR #102, opened 2021-05-10
post employee/vendor optimisation

NileshPant1999:


apps/netsuite/utils.py
```diff
@@ -439,6 +430,49 @@ def sync_employees(self):
...
+    def get_or_create_vendor(self, expense_attribute: ExpenseAttribute, expense_group: ExpenseGroup):
+        vendor = self.connection.vendors.search(
+            attribute='entityId', value=expense_attribute.detail['full_name'], operator='is')
+
+        if not vendor:
+            created_vendor = self.post_vendor(expense_attribute, expense_group)
+            return self.create_destination_attribute(
+                'vendor', expense_attribute.detail['full_name'], created_vendor['internalId'], expense_attribute.value)
+        else:
+            vendor = vendor[0]
```
ashwin1111, line 455: Does `self.connection.vendors.search` return a list every time? Ex: When you have 1 employee with matching result or > 1 employee with matching result

apps/netsuite/utils.py
```diff
@@ -562,16 +596,6 @@ def post_employee(self, employee: ExpenseAttribute, expense_group: ExpenseGroup)
         }
         created_employee = self.connection.employees.post(employee)
```
ashwin1111, line 597: `return self.connection.employees.post(employee)`

ashwin1111:
Also, fix the PR title

apps/netsuite/utils.py
NileshPant1999: self.connection.vendors.search() will always return 1 it is not possible to create two vedors with same entityid's

ashwin1111 (approved):

apps/netsuite/utils.py
Shwetabhk: @NileshPant1999 I think we can have 2 vendors with the same entityId but yeah even if the result returns only 1 vendor it will be in a list

Shwetabhk (approved):

PR #104, opened 2021-05-13
Feature: Adding support for credit card charges in NetSuite

Shwetabhk:


apps/netsuite/models.py
```diff
@@ -383,6 +385,169 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
...
+    created_at = models.DateTimeField(auto_now_add=True, help_text='Created at')
+    updated_at = models.DateTimeField(auto_now=True, help_text='Updated at')
+
+    class Meta:
+        db_table = 'credit_card_charges'
+
+    @staticmethod
+    def create_charge_card_transaction(expense_group: ExpenseGroup):
+        """
+        Create Charge Card Transaction
```
ashwin1111, line 411: Can we try to maintain same term everywhere. `NetSuite Credit Card Charge`

apps/netsuite/models.py
```diff
@@ -383,6 +385,169 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
...
+    external_id = models.CharField(max_length=255, unique=True, help_text='Fyle reimbursement id')
+    transaction_date = models.DateTimeField(help_text='CC Charge transaction date')
+    created_at = models.DateTimeField(auto_now_add=True, help_text='Created at')
+    updated_at = models.DateTimeField(auto_now=True, help_text='Updated at')
+
+    class Meta:
+        db_table = 'credit_card_charges'
+
+    @staticmethod
+    def create_charge_card_transaction(expense_group: ExpenseGroup):
```
ashwin1111, line 409: Can we try to maintain same term everywhere. `create_charge_card_charge`

apps/netsuite/models.py
```diff
@@ -383,6 +385,169 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
...
+
+        expense_group.description['spent_at'] = expense.spent_at.strftime("%Y-%m-%d")
+        expense_group.save()
+
+        if not vendor:
+            vendor_id = general_mappings.default_ccc_vendor_id
+        else:
+            vendor_id = vendor.destination_id
+
+        credit_charge_object, _ = CreditCardCharge.objects.update_or_create(
```
ashwin1111, line 440: can we do just create()? instead of update_or_create().
Since this is inside a transaction block, I think create() won't cause an issue.

apps/netsuite/models.py
```diff
@@ -383,6 +385,169 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
...
+        """
+        expenses = expense_group.expenses.all()
+        credit_card_charge = CreditCardCharge.objects.get(expense_group=expense_group)
+        general_settings: WorkspaceGeneralSettings = WorkspaceGeneralSettings.objects.get(
+            workspace_id=expense_group.workspace_id)
+        general_mappings = GeneralMapping.objects.get(workspace_id=expense_group.workspace_id)
+
+        credit_card_charge_lineitem_objects = []
+
+        for lineitem in expenses:
```
ashwin1111, line 495: Since line items are always going to be 1, do we need a loop?

apps/netsuite/models.py
```diff
@@ -383,6 +385,169 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
...
+        db_table = 'credit_card_charge_lineitems'
+
+    @staticmethod
+    def create_credit_card_charge_lineitems(expense_group: ExpenseGroup):
+        """
+        Create credit card charge lineitems
+        :param expense_group: expense group
+        :return: credit card charge lineitems objects
+        """
+        expenses = expense_group.expenses.all()
```
ashwin1111, line 487: Since line items are always going to be 1, can we do .first()?

apps/netsuite/models.py
```diff
@@ -383,6 +385,169 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
...
+                customer_id = get_customer_id_or_none(expense_group, lineitem)
+
+            billable = lineitem.billable
+            if customer_id:
+                if not billable:
+                    billable = False
+            else:
+                billable = None
+
+            credit_card_charge_lineitem_object, _ = CreditCardChargeLineItem.objects.update_or_create(
```
ashwin1111, line 529: can we do just create()

apps/netsuite/models.py
Shwetabhk: sounds fair, will do

apps/netsuite/models.py
Shwetabhk: Yes

apps/netsuite/models.py
Shwetabhk: update_or_create acts as an extra safety net in this case, since it directly relates to export, I would say lets hold off on doing this for a while

apps/netsuite/utils.py
```diff
@@ -868,9 +877,147 @@ def get_bill(self, internal_id):
...
+    def __construct_credit_card_charge_lineitems(
+            credit_card_charge_lineitems: List[CreditCardChargeLineItem],
+            attachment_links: Dict, cluster_domain: str, org_id: str) -> List[Dict]:
+        """
+        Create credit_card_charge line items
+        :return: constructed line items
+        """
+        lines = []
+
+        for line in credit_card_charge_lineitems:
```
ashwin1111, line 890: same

apps/netsuite/models.py
Shwetabhk: right, removing loop

apps/workspaces/utils.py
```diff
@@ -43,6 +44,16 @@ def create_or_update_general_settings(general_settings_payload: Dict, workspace_
         }
     )
 
+    if general_settings.map_merchant_to_vendor and \
+            general_settings.corporate_credit_card_expenses_object == 'CREDIT CARD CHARGE':
+        expense_group_settings = ExpenseGroupSettings.objects.get(workspace_id=workspace_id)
+
+        ccc_expense_group_fields = expense_group_settings.corporate_credit_card_expense_group_fields
+        ccc_expense_group_fields.append('expense_id')
+        expense_group_settings.corporate_credit_card_expense_group_fields = ccc_expense_group_fields
```
ashwin1111, line 53: remove duplicates. If `expense_id` already exists, we end up appending twice.

apps/netsuite/utils.py
```diff
@@ -77,7 +83,10 @@ def sync_accounts(self):
                         'attribute_type': 'CREDIT_CARD_ACCOUNT',
                         'display_name': 'Credit Card Account',
                         'value': account['acctName'],
-                        'destination_id': account['internalId']
+                        'destination_id': account['internalId'],
+                        'detail': {
+                            'account_type': account['accType']
```
Shwetabhk, line 88: @ashwin1111 We'll need to filter type `_creditCard` for Credit Card Transactions on frontend

ashwin1111: @Shwetabhk typo, `account['acctType']`

apps/netsuite/tasks.py
```diff
@@ -571,12 +692,48 @@ def schedule_bills_creation(workspace_id: int, expense_group_ids: List[str]):
...
+    Schedule Credit Card Charge creation
+    :param expense_group_ids: List of expense group ids
+    :param workspace_id: workspace id
+    :return: None
+    """
+    if expense_group_ids:
+        expense_groups = ExpenseGroup.objects.filter(
+            Q(tasklog__id__isnull=True) | ~Q(tasklog__status__in=['IN_PROGRESS', 'COMPLETE']),
+            workspace_id=workspace_id, id__in=expense_group_ids,
+            credit_card_charge__id__isnull=True, exported_at__isnull=True
```
ashwin1111, line 706: `creditcardcharge`

NileshPant1999 (approved):

PR #115, opened 2021-06-03
default department in case when ccc is set to bill

NileshPant1999:


apps/netsuite/models.py
```diff
@@ -372,7 +372,7 @@ def create_bill_lineitems(expense_group: ExpenseGroup):
                     'location_id': location_id if general_mappings.location_level in [
                         'TRANSACTION_LINE', 'ALL'] else None,
                     'class_id': class_id,
-                    'department_id': department_id,
+                    'department_id': department_id if department_id else general_mappings.department_id,
```
ashwin1111, line 375: Don't we need a ccc check?

NileshPant1999: yes, we need to only add department_id when ccc_config is bill.

NileshPant1999: okay will add that

PR #131, opened 2021-07-09
adding basic test setup

Sravanksk:


fyle_netsuite_api/tests/helpers.py
```diff
@@ -0,0 +1,69 @@
...
+        self.auth_token.save()
+
+        return fyle_connection
+
+    def get_user_workspace_id(self):
+        """
+        GET workspace_id of the user
+        """
+        self.client.credentials(HTTP_AUTHORIZATION='Bearer ' + self.access_token)
+        self.client.post('{0}/workspaces/'.format(settings.API_URL),
```
Shwetabhk, line 66: Why are we still calling this here?

apps/mappings/tests/test_views.py
```diff
@@ -0,0 +1,43 @@
...
+
+        self.client = APIClient()
+        test_helpers.client = self.client
+        self.workspace = test_helpers.get_user_workspace_id()
+
+        self.__headers = {
+            'Authorization': 'Bearer {}'.format(access_token)
+        }
+
+        self.subsidiary_mappings_payload = {
```
Shwetabhk, line 24: Let's add payloads in a separate file, `fixtures.py`

fyle_netsuite_api/tests/helpers.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+        self.auth_token.save()
+
+        return fyle_connection
+
+    def get_user_workspace_id(self):
+        """
+        GET workspace_id of the user
+        """
+        self.client.credentials(HTTP_AUTHORIZATION='Bearer ' + self.access_token)
```
Shwetabhk, line 65: use string formatting
`'Bearer {0}'.format(self.access_token)`

fyle_netsuite_api/tests/helpers.py
Shwetabhk: We can discuss how you can call this

PR #133, opened 2021-07-12
move netsuite objects sync to generic function

Sravanksk:


apps/mappings/tasks.py
```diff
@@ -456,6 +450,32 @@ def schedule_auto_map_ccc_employees(workspace_id: int):
             schedule.delete()
 
 
+def sync_netsuite_attribute(netsuite_attribute_type, workspace_id):
```
ashwin1111, line 453: type pliss

ashwin1111 (approved):

apps/mappings/tasks.py
Sravanksk: 😄

PR #134, opened 2021-07-12
Custom Segments cleanup

ashwin1111:


apps/netsuite/signals.py
```diff
@@ -0,0 +1,51 @@
+"""
+NetSuite Signals
+"""
+import traceback
+
+from rest_framework.response import Response
```
ashwin1111, line 6: TODO: remove unused imports

Shwetabhk (approved):

PR #137, opened 2021-07-15
NetSuite API: Tasks App Cleanup - Rename, Refactor, Remove, Documentation

ashwin1111:


apps/fyle/views.py
```diff
@@ -63,19 +63,13 @@ def get(self, request, *args, **kwargs):
         )
 
 
-class ExpenseGroupSettingsView(generics.ListCreateAPIView):
+class ExpenseGroupSettingsView(generics.RetrieveAPIView):
```
Shwetabhk, line 66: use a parent class with post method allowed

Shwetabhk (approved):

PR #157, opened 2021-09-22
POST attachments to a single folder in NetSuite

Sravanksk:


apps/netsuite/tasks.py
```diff
@@ -38,16 +38,16 @@ def load_attachments(netsuite_connection: NetSuiteConnector, expense_id: str, ex
     :param expense_group: Integration Expense group
     """
     workspace_id = expense_group.workspace_id
+    workspace = Workspace.objects.get(id=workspace_id)
```
ashwin1111, line 41: `expense_group.workspace` would give workspace instance, this is not needed

ashwin1111 (approved):

Shwetabhk (approved):

PR #168, opened 2021-10-12
Fetching Tax Group along with codes

NileshPant1999:


apps/netsuite/connector.py
```diff
@@ -601,6 +603,24 @@ def sync_tax_items(self):
                                 'tax_rate': tax_rate
                             }
                         })
+
+            for tax_groups in tax_groups_generator:
```
ashwin1111, line 607: is this supposed to be inside `tax_items_generator` loop?

NileshPant1999: no i will fix this didn't get caught on testing as it was updating it again and again i think

Sravanksk (approved):

PR #212, opened 2022-01-27
Export refunds from Personal cards to Netsuite Expense report

NileshPant1999:


apps/fyle/models.py
```diff
@@ -308,14 +308,33 @@ def create_expense_groups_by_report_id_fund_source(expense_objects: List[Expense
         Group expense by report_id and fund_source
         """
         expense_group_settings = ExpenseGroupSettings.objects.get(workspace_id=workspace_id)
+        configuration = Configuration.objects.get(workspace_id=workspace_id)
```
ashwin1111, line 311: we have configurations available in parent function, can you plis make use of them

ashwin1111 (approved):

Sravanksk (approved):

PR #222, opened 2022-02-10
Credit card mapping support for netsuite

NileshPant1999:


apps/netsuite/models.py
```diff
@@ -247,6 +247,29 @@ def get_expense_purpose(lineitem, category, configuration) -> str:
 
     return purpose
 
+def get_ccc_account_id(configuration, general_mappings, expense, description):
```
ashwin1111, line 250: types please

apps/workspaces/models.py
```diff
@@ -1,6 +1,7 @@
 """
 Workspace Models
 """
+from email.policy import default
```
ashwin1111, line 4: arey bhaai, what is this?

ashwin1111 (approved):

Sravanksk (approved):

PR #245, opened 2022-03-16
Email Notification template and bug fix

NileshPant1999:


requirements.txt
```diff
@@ -58,3 +57,4 @@ urllib3==1.26.5
 wcwidth==0.1.8
 wrapt==1.12.1
 zeep==4.1.0
+django-sendgrid-v5
```
Shwetabhk, line 60: can we add the specific version that is installed?
It this gets a major version upgrade things will start breaking, also please order this alphabetically

Shwetabhk (approved):

PR #246, opened 2022-03-18
post NS vendors as merchants to fyle

labhvam5:


apps/mappings/tasks.py
```diff
@@ -13,11 +13,13 @@
     CategoryMapping
 from fyle_accounting_mappings.helpers import EmployeesAutoMappingHelper
 
+#todo will update the import 
 from apps.fyle.connector import FyleConnector
-from fyle_integrations_platform_connector import PlatformConnector
+# from fyle_integrations_platform_connector import PlatformConnector
```
Sravanksk, line 18: can remove this

apps/mappings/tasks.py
```diff
@@ -871,3 +876,85 @@ def schedule_fyle_attributes_creation(workspace_id: int):
...
+def auto_create_vendors_as_merchants(workspace_id):
+    try:
+        fyle_credentials: FyleCredential = FyleCredential.objects.get(workspace_id=workspace_id)
+
+        fyle_connection = PlatformConnector(fyle_credentials)
+
+        existing_merchants_name = ExpenseAttribute.objects.filter(attribute_type='MERCHANT', workspace_id=workspace_id)
+        first_run = True
+        if existing_merchants_name:
+            first_run = False
```
Sravanksk, line 920: can do this `first_run = True if existing_merchants_name else False`

labhvam5: very good suggestion 🤯

.gitignore
```diff
@@ -154,3 +154,7 @@ cache.db-journal
 # Test scripts
 test_scripts/
 fyle_accounting_mappings/
+
+#todo will remove this off 
+#connector
+apps/connector
```
Shwetabhk, line 160: do we need to remove this?

apps/mappings/tasks.py
```diff
@@ -871,3 +876,83 @@ def schedule_fyle_attributes_creation(workspace_id: int):
...
+    
+    if first_run:
+        netsuite_attributes = DestinationAttribute.objects.filter(
+            attribute_type='VENDOR', workspace_id=workspace_id).order_by('value', 'id')
+    else:
+        merchant = platform_connection.merchants.get()
+        merchant_updated_at = parser.isoparse(merchant['updated_at']).strftime('%Y-%m-%d')
+        today_date = datetime.now().strftime('%Y-%m-%d')
+        netsuite_attributes = DestinationAttribute.objects.filter(
+            attribute_type='VENDOR', workspace_id=workspace_id, updated_at__range=[merchant_updated_at,today_date]).order_by('value', 'id')
```
Shwetabhk, line 899: this line is too long break it

Shwetabhk:
```suggestion
            attribute_type='VENDOR', workspace_id=workspace_id, 
            updated_at__range=[merchant_updated_at, today_date]).order_by('value', 'id')
```

Shwetabhk (approved):

.gitignore
labhvam5: not now but eventually yes will keep this in mind

Shwetabhk: can let it be

PR #252, opened 2022-04-08
Adding patch for when only customer is present and projects are not p…

Shwetabhk:
…resent

### DIFF
--- a/apps/netsuite/views.py
+++ b/apps/netsuite/views.py
@@ -88,7 +88,7 @@ class NetSuiteFieldsView(generics.ListAPIView):
         for attribute in attributes:
             if attribute['display_name'] == 'Project':
                 has_projects = True
-            elif attribute['display_name'] == 'Project':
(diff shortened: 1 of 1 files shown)


apps/netsuite/views.py
```diff
@@ -79,10 +79,23 @@ def get_queryset(self):
...
+            ~Q(attribute_type='CCC_ACCOUNT') & ~Q(attribute_type='TAX_ITEM'),
             workspace_id=self.kwargs['workspace_id']
         ).values('attribute_type', 'display_name').distinct()
 
+        has_projects = False
+        has_customers = False
+        for attribute in attributes:
+            if attribute['display_name'] == 'Project':
+                has_projects = True
+            elif attribute['display_name'] == 'Project':
```
Shwetabhk, line 91: bug

Sravanksk: `elif attribute['display_name'] == 'Customer':`

Sravanksk (approved):

PR #255, opened 2022-04-08
filtering customer even if project is present

Shwetabhk:


apps/netsuite/views.py
```diff
@@ -99,8 +99,7 @@ def get_queryset(self):
                 'display_name': 'Project'
             })
         
-        if has_customers and has_projects:
-            attributes = list(filter(lambda attribute: attribute['display_name'] != 'Customer', attributes))
+        attributes = list(filter(lambda attribute: attribute['display_name'] != 'Customer', attributes))
```
Shwetabhk, line 102: We are append `Project` above anyway
so removing Customer

Sravanksk (approved):

PR #257, opened 2022-04-22
append NetSuite parent class name to sub-classes

Sravanksk:


apps/netsuite/connector.py
```diff
@@ -311,10 +311,14 @@ def sync_classifications(self):
         classification_attributes = []
 
         for classification in classifications:
+            value = classification['name']
+            if classification['parent']:
```
ashwin1111, line 315: Shouldn't we check for `classification['parent']['name']` as well?

ashwin1111 (approved):

apps/netsuite/connector.py
Sravanksk: This is not needed - please check this API response 

If parent exists - 

```json
 "parent": {
        "name": "AZ",
        "internalId": "14",
        "externalId": "None",
        "type": "None"
    }
  ```
  If parent does not exist 
  ```json
   "parent": "None"

NileshPant1999 (approved):

PR #262, opened 2022-05-03
Phase 4 Migrations (Post API's)

NileshPant1999:


apps/mappings/tasks.py
```diff
@@ -55,8 +55,10 @@ def create_fyle_categories_payload(categories: List[DestinationAttribute], works
         if category.value not in existing_category_names:
             payload.append({
                 'name': category.value,
+                'sub_category': ' ',
```
ashwin1111, line 58: why `' '` and not `''`

Sravanksk, line 58: this is not needed now, pls remove

Sravanksk (approved):

PR #280, opened 2022-06-16
Update changes in accounting software for projects 

labhvam5:


apps/mappings/tasks.py
```diff
@@ -419,26 +419,68 @@ def create_fyle_tax_group_payload(netsuite_attributes: List[DestinationAttribute
...
+		active=False,
+		workspace_id=workspace_id
+	)
+
+    destination_attributes_without_duplicates = remove_duplicates(destination_attributes)
+    destination_attribute_ids = [destination_attribute.id for destination_attribute in destination_attributes_without_duplicates]
+
+	# Get all the expense attributes that are mapped to these destination_attribute_ids
+    expense_attributes = ExpenseAttribute.objects.filter(
+		attribute_type=mapping_setting.source_field,
```
ashwin1111, line 439: indent

apps/mappings/tasks.py
```diff
@@ -419,26 +419,68 @@ def create_fyle_tax_group_payload(netsuite_attributes: List[DestinationAttribute
...
+	# Get All the inactive destination attribute ids
+    destination_attributes = DestinationAttribute.objects.filter(
+		attribute_type=mapping_setting.destination_field, 
+		mapping__isnull=False,
+		mapping__destination_type=mapping_setting.destination_field,
+		active=False,
+		workspace_id=workspace_id
+	)
+
+    destination_attributes_without_duplicates = remove_duplicates(destination_attributes)
```
ashwin1111, line 434: there cannot be any duplicates here since we add mapping__isnull=False

apps/mappings/tasks.py
```diff
@@ -464,6 +506,15 @@ def post_projects_in_batches(platform: PlatformConnector, workspace_id: int, des
             platform.projects.sync()
 
         Mapping.bulk_create_mappings(paginated_ns_attributes, 'PROJECT', destination_field, workspace_id)
+    
+    mapping_setting = MappingSetting.objects.filter(workspace_id=workspace_id, destination_field='PROJECT').first()
+    projects_to_be_disabled_ids = disable_inactive_expense_attributes(mapping_setting, workspace_id)
+
+    if projects_to_be_disabled_ids:
+        expense_attributes = ExpenseAttribute.objects.filter(id__in = projects_to_be_disabled_ids)
```
ashwin1111, line 514: no spacing, `id__in=projects_to_be_disabled_ids`

apps/mappings/tasks.py
```diff
@@ -464,6 +506,15 @@ def post_projects_in_batches(platform: PlatformConnector, workspace_id: int, des
...
+        fyle_payload: List[Dict] = create_fyle_projects_payload(projects= [], existing_project_names= [],  updated_projects=expense_attributes)
```
Sravanksk, line 515: indentation

apps/mappings/tasks.py
```diff
@@ -419,26 +419,68 @@ def create_fyle_tax_group_payload(netsuite_attributes: List[DestinationAttribute
...
                     project.value,
-                    project.destination_id
+                    destination_id_of_project
                 ),
-                'is_enabled': True if project.active is None else project.active
+                'is_enabled': False
             })
+    else:
+        for project in projects:
+            if project.value not in existing_project_names:
```
ashwin1111, line 474: let's also add a check for project.active here. Else we might bring in all disabled projects from accounting software and keep it disabled in Fyle. Consider an org that takes up 100 projects every year and once completed marks them as inactive. If the org completes 3 years, then we might end up adding 300 disabled projects to Fyle (in a disabled state) which is not needed

ashwin1111 (approved):

Sravanksk (approved):

PR #302, opened 2022-09-16
fix sync of tax_groups

Sravanksk:


apps/netsuite/connector.py
```diff
@@ -66,7 +66,10 @@ def get_message_and_code(raw_response):
     
     @staticmethod
     def get_tax_code_name(item_id, tax_type, rate):
-        return '{0}: {1} @{2}%'.format(tax_type, item_id, rate)
+        if tax_type and rate:
+            return '{0}: {1} @{2}%'.format(tax_type, item_id, rate)
+        else:
+            return '{0} @{1}%'.format(item_id, rate)
```
Shwetabhk, line 72: will rate always be available?

Shwetabhk (approved):

apps/netsuite/connector.py
NileshPant1999: yes @Shwetabhk if the value is not there we pass 0 in the function

PR #313, opened 2022-11-21
Import netsuite employees to fyle

harshithaputtaswamy:


ashwin1111:
Will review it after @NileshPant1999's green

apps/mappings/tasks.py
```diff
@@ -1030,3 +1033,154 @@ def schedule_vendors_as_merchants_creation(import_vendors_as_merchants, workspac
...
+
+    if departments:
+        for department in departments:
+            if not department['is_enabled']:
+                return {
+                    'name': department_name,
+                    'id': department['id'],
+                    'is_enabled': True
+                }
+    else:
```
NileshPant1999, line 1057: identation is wrong, ir should go with other if i think

apps/mappings/tasks.py
```diff
@@ -1030,3 +1033,154 @@ def schedule_vendors_as_merchants_creation(import_vendors_as_merchants, workspac
...
+    })
+
+    departments = []
+
+    for response in departments_generator:
+        if response.get('data'):
+            departments.extend(response['data'])
+
+    if departments:
+        for department in departments:
```
NileshPant1999, line 1050: also can you explain what are you trying to do here

apps/mappings/tasks.py
harshithaputtaswamy: First if is to check if department exists and update department is_enables to True if it was disabled before
The control will enter else if there is no department found and create a new dept

apps/netsuite/connector.py
```diff
@@ -1,3 +1,4 @@
+import re
```
NileshPant1999, line 1: is this getting used anywhere

apps/mappings/tasks.py
```diff
@@ -1030,3 +1033,154 @@ def schedule_vendors_as_merchants_creation(import_vendors_as_merchants, workspac
...
+
+
+def create_fyle_department_payload(platform_connection: PlatformConnector, department_name: str):
+    departments_generator = platform_connection.departments.search_departments(query_params={
+        'name': 'eq.{}'.format(department_name)
+    })
+
+    departments = []
+
+    for response in departments_generator:
```
NileshPant1999, line 1045: can add a check directly in the generator

NileshPant1999 (approved):

apps/mappings/tasks.py
```diff
@@ -1030,3 +1033,150 @@ def schedule_vendors_as_merchants_creation(import_vendors_as_merchants, workspac
...
+            platform_connection.departments.post(department)
+
+    if fyle_employee_payload:
+        platform_connection.connection.v1beta.admin.employees.invite_bulk({'data': fyle_employee_payload})
+
+        workspace.employee_exported_at = datetime.now()
+        workspace.save()
+
+    if employee_approver_payload:
+        platform_connection.connection.v1beta.admin.employees.invite_bulk({'data': employee_approver_payload})
```
ashwin1111, line 1125: what happens when this approver is not a part of Fyle? will the user get an invite?

apps/netsuite/connector.py
```diff
@@ -465,28 +465,43 @@ def sync_employees(self):
...
                 detail = {
                     'email': employee['email'] if employee['email'] else None,
                     'department_id': employee['department']['internalId'] if employee['department'] else None,
                     'location_id': employee['location']['internalId'] if employee['location'] else None,
-                    'class_id': employee['class']['internalId'] if employee['class'] else None
+                    'class_id': employee['class']['internalId'] if employee['class'] else None,
+                    'department_name': employee['department']['name'] if employee['department'] else None,
+                    'location_name': employee['location']['name'] if employee['location'] else None,
+                    'full_name': ' '.join(filter(None, [employee['firstName'], employee['middleName'], employee['lastName']])),
+                    'joined_at': employee['dateCreated'].isoformat(timespec='milliseconds'),
```
ashwin1111, line 481: handle safely, if employee['dateCreated']

apps/workspaces/models.py
```diff
@@ -27,6 +27,7 @@ class Workspace(models.Model):
     destination_synced_at = models.DateTimeField(help_text='Datetime when destination dimensions were pulled', null=True)
     created_at = models.DateTimeField(auto_now_add=True, help_text='Created at datetime')
     updated_at = models.DateTimeField(auto_now=True, help_text='Updated at datetime')
+    employee_exported_at = models.DateTimeField(auto_now_add=True, help_text='Employee exported to Fyle at datetime')
```
ashwin1111, line 30: Why is this auto_now_add=True?

apps/mappings/tasks.py
```diff
@@ -1030,3 +1033,150 @@ def schedule_vendors_as_merchants_creation(import_vendors_as_merchants, workspac
...
+        if employee.value not in existing_employee_names:
+            if employee.detail["department_name"]:
+                department = create_fyle_department_payload(platform_connection, employee.detail["department_name"])
+                if department:
+                    department_payload.extend(department)
+
+            employee_payload.append({
+                "user_email": employee.detail["email"],
+                "user_full_name": employee.detail["full_name"],
+                "code": employee.destination_id,
```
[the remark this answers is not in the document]
harshithaputtaswamy: Had confirmed with Kamalini, the customer is using internal_id in Netsuite as unique identifier for employees

apps/workspaces/models.py
harshithaputtaswamy: It should not get updated when we do .save() elsewhere other than in the post_employees() call

apps/netsuite/connector.py
harshithaputtaswamy: It's an auto generated field in Netsuite, do we still have to have a check for it? @ashwin1111

apps/mappings/tasks.py
harshithaputtaswamy: Should check this part

harshithaputtaswamy: Approver will also be part of Fyle no, if approver is not added to fyle then the employee post call throws error

ashwin1111: How do you say approver exists in Fyle already? shouldn't we handle this safely?

ashwin1111 (approved):

ashwin1111 (approved):

PR #314, opened 2022-11-30
simply report closure

labhvam5:


apps/fyle/models.py
```diff
@@ -37,6 +37,12 @@
     'PERSONAL_CORPORATE_CREDIT_CARD_ACCOUNT': 'CCC'
 }
 
+CCC_EXPENSE_STATE = (
+    ('APPROVED', 'APPROVED'),
+    ('PAID', 'PAID'),
+    ('PAYMENT_PROCESSING', 'PAYMENT_PROCESSING')
```
ashwin1111, line 43: This is to support existing customers?

apps/fyle/tasks.py
```diff
@@ -70,17 +71,35 @@ def create_expense_groups(workspace_id: int, configuration: Configuration, fund_
...
+                    last_paid_at=last_synced_at if expense_group_settings.expense_state == 'PAID' else None
+                ))
 
             if expenses:
                 workspace.last_synced_at = datetime.now()
-                workspace.save()
+                reimbursable_expenses_count = len(expenses)
+
+            if 'CCC' in fund_source:
+                expenses.extend(platform.expenses.get(
```
ashwin1111, line 91: what happens for existing customers? those who have `PAYMENT_PROCESSING` state.

apps/netsuite/views.py
```diff
@@ -174,7 +174,7 @@ def post(self, request, *args, **kwargs):
                 },
                 status=status.HTTP_400_BAD_REQUEST
             )
-        except Exception : 
+        except Exception :
```
ashwin1111, line 177: `except Exception:`

sql/scripts/016-set-ccc-expense-state.sql
```diff
@@ -0,0 +1,16 @@
+rollback;
+begin;
+
+update 
+  expense_group_settings 
+set 
+  ccc_expense_state = expense_group_settings.expense_state 
+where
```
ashwin1111, line 8: where is not required

sql/scripts/017-fill-ccc-last-synced-at.sql
```diff
@@ -0,0 +1,9 @@
+rollback;
+begin;
+
+update 
+  workspaces 
+set 
+  ccc_last_synced_at = last_synced_at 
+where
```
ashwin1111, line 8: this is not required

apps/fyle/models.py
labhvam5: yes.

apps/fyle/tasks.py
labhvam5: we need to add settled_at for the ccc expense import call .

Shwetabhk (approved):

PR #321, opened 2022-12-09
auto-sync coa and allow force disable coa, projects

harshithaputtaswamy:
Changes implemented:

1. modified sync_accounts and sync_expense_categories - On the first run get only the active entities from netsuite, update the active flag on subsequent sync runs
2. Allow users to force disable projects from fyle, integrations will not enable it on fyle even when it is enabled on netsuite
3. Auto-sync categories and accounts - update active column based on value in netsuite
4. If the user disables a category or chart of account on netsuite then the same will be reflected on fyle
5. Allow users to force disable categories from fyle, integrations will not enable it on fyle even when it is enabled on netsuite


apps/mappings/tasks.py
```diff
@@ -37,6 +37,75 @@ def remove_duplicates(ns_attributes: List[DestinationAttribute]):
     return unique_attributes
 
 
+def disable_or_enable_expense_attributes(source_field, destination_field, workspace_id):
+
+    # Get All the inactive destination attribute ids
+    filter = {
+        'mapping__isnull': False,
+		'mapping__destination_type': destination_field
```
NileshPant1999, line 45: remove extra spacing

tests/test_mappings/test_tasks.py
```diff
@@ -820,3 +851,10 @@ def test_auto_create_netsuite_employees_on_fyle(db, mocker):
     fyle_credentials.delete()
 
     auto_create_netsuite_employees_on_fyle(workspace_id=workspace_id)
+
+
+def test_disable_or_enable_expense_attributes(db):
+    workspace_id = 2
```
NileshPant1999, line 857: can directly pass this no ?? it is only getting used once

tests/test_mappings/test_tasks.py
```diff
@@ -347,19 +378,19 @@ def test_auto_create_project_mappings(db, mocker):
...
         response = auto_create_project_mappings(workspace_id=1)
 
         mock_call.side_effect = Exception()
         response = auto_create_project_mappings(workspace_id=1)
 
+    fyle_credentials = FyleCredential.objects.get(workspace_id=workspace_id)
+    fyle_credentials.delete()
+
+    response = auto_create_project_mappings(workspace_id=workspace_id)
+    assert response == None
```
NileshPant1999, line 392: can we actually have some mappings here, i mean instead of none can do later too in a separate pr

apps/mappings/tasks.py
```diff
@@ -58,7 +127,7 @@ def get_all_categories_from_fyle(platform: PlatformConnector):
     return category_name_map
 
 
-def create_fyle_categories_payload(categories: List[DestinationAttribute], category_map: Dict):
+def create_fyle_categories_payload(categories: List[DestinationAttribute], category_map: Dict, updated_categories: List[ExpenseAttribute] = [], destination_type: str = ''):
```
NileshPant1999, line 130: can set `destination_type` to null

tests/test_mappings/test_tasks.py
harshithaputtaswamy: This is to check the exception where fyle_creds don't exsist

tests/test_mappings/test_tasks.py
harshithaputtaswamy: For future use, if some more things are added to this function

apps/mappings/tasks.py
```diff
@@ -37,6 +37,75 @@ def remove_duplicates(ns_attributes: List[DestinationAttribute]):
...
+                'categorymapping__destination_account_id__in': destination_attribute_ids
+            }
+
+    expense_attributes_to_disable = ExpenseAttribute.objects.filter(
+        attribute_type=source_field,
+        active=True,
+        **filter
+    )
+
+    expense_attributes_to_enable = []
```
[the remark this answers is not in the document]
harshithaputtaswamy: That is when source is not project or category, wanted to return an empty list otherwise

ashwin1111:
please assign back review once the action items are covered (https://fylein.slack.com/archives/C0472NQ5FFW/p1670941298992389?thread_ts=1670939099.821389&cid=C0472NQ5FFW)

apps/netsuite/connector.py
```diff
@@ -87,14 +87,17 @@ def sync_accounts(self):
                 'ccc_account': [],
                 'vendor_payment_account': []
             }
+            destination_ids = DestinationAttribute.objects.filter(workspace_id=self.workspace_id,
+                attribute_type= 'ACCOUNT', display_name='Account').values_list('destination_id', flat=True)
```
ashwin1111, line 91: `attribute_type='ACCOUNT'`

ashwin1111 (approved):

PR #325, opened 2022-12-27
reset workspace function commit

DhaaraniCIT:


apps/workspaces/permissions.py
```diff
@@ -31,3 +31,19 @@ def has_permission(self, request, view):
...
+class IsAuthenticatedForTest(permissions.BasePermission):
+    """
+    Custom auth for preparing a workspace for e2e tests
+    """
+    def has_permission(self, request, view):
+        # Client sends a token in the header, which we decrypt and compare with the Client Secret
+        cipher_suite = Fernet(settings.ENCRYPTION_KEY)
+        try:
+            decrypted_password = cipher_suite.decrypt(request.headers['X-E2E-Tests-Client-ID'].encode('utf-8')).decode('utf-8')
+            if decrypted_password == settings.E2E_TESTS_CLIENT_SECRET:
```
ashwin1111, line 44: we'll need to add E2E_TESTS_CLIENT_SECRET in settings.py, that's missing

apps/workspaces/permissions.py
```diff
@@ -31,3 +31,19 @@ def has_permission(self, request, view):
...
             workspace_users = Workspace.objects.filter(pk=workspace_id).values_list('user', flat=True)
             return self.validate_and_cache(workspace_users, user, workspace_id, True)
+
+class IsAuthenticatedForTest(permissions.BasePermission):
+    """
+    Custom auth for preparing a workspace for e2e tests
+    """
+    def has_permission(self, request, view):
+        # Client sends a token in the header, which we decrypt and compare with the Client Secret
+        cipher_suite = Fernet(settings.ENCRYPTION_KEY)
```
ashwin1111, line 41: same as line 44. Also, add the value in Integrations Central and fyle-deploy too.

ashwin1111: can copy same value on what is there for QBO

apps/workspaces/views.py
```diff
@@ -456,3 +457,32 @@ def get(self, request, *args, **kwargs):
...
+
+            # Filter out prod orgs
+            if 'fyle for' in workspace.name.lower():
+                with transaction.atomic():
+                    # Reset the workspace completely
+                    with connection.cursor() as cursor:
+                        cursor.execute('select reset_workspace(%s)', [workspace.id])
+
+                error_message = 'Not e2e test worksoace'
+            return Response(status=status.HTTP_400_BAD_REQUEST, data={'message': error_message})
```
ashwin1111, line 484: return 200, remove error_message, just send empty dict {}

ashwin1111 (approved):

PR #335, opened 2023-01-10
Bug Fix: Add Check for Accounts while Create CCC Mappings

NileshPant1999:


apps/mappings/tasks.py
```diff
@@ -233,7 +233,10 @@ def bulk_create_ccc_category_mappings(workspace_id: int):
     mapping_updation_batch = []
 
     for category_mapping in category_mappings:
-        ccc_account_id = destination_id_pk_map[category_mapping.destination_expense_head.detail['account_internal_id'].lower()]
+        # Adding a check if the account_internal_id has a account accociated in destination attributes table
+        ccc_account_id = destination_id_pk_map[category_mapping.destination_expense_head.detail['account_internal_id'].lower()] \
+            if category_mapping.destination_expense_head.detail['account_internal_id'].lower() in destination_id_pk_map else None        
+
         mapping_updation_batch.append(
```
Shwetabhk, line 240:
```suggestion
        if ccc_account_id:
            mapping_updation_batch.append(
```

PR #347, opened 2023-02-02
Serve application using gunicorn

ashwin1111:
https://www.notion.so/fyleuniverse/Fix-Broken-Windows-Backend-Bring-Fatal-Tasks-to-0-and-clean-up-integrations_logs_prod-face9bc3d06c40a49889fa2fdf1d9313#132a015e989c4bf6898f250ba640fba6

### DIFF
--- a/deployment/staging/controller.yml
+++ b/deployment/staging/controller.yml
@@ -36,9 +36,9 @@ spec:
           image: docker.io/{{DOCKERHUB_USERNAME}}/fyle_netsuite-api:{{RELEASE_VERSION}}
           resources:
             limits:
-              memory: 600Mi
+              memory: 450Mi
             requests:
-              memory: 600Mi
+              memory: 450Mi
           env:
             - name: GET_HOSTS_FROM
               value: dns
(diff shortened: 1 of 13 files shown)


deployment/staging/controller.yml
```diff
@@ -36,9 +36,9 @@ spec:
           image: docker.io/{{DOCKERHUB_USERNAME}}/fyle_netsuite-api:{{RELEASE_VERSION}}
           resources:
             limits:
-              memory: 300Mi
+              memory: 600Mi
```
Shwetabhk, line 39: 600 mb is too high, lets reduce it to 300 only
we can reduce workers

deployment/staging/controller.yml
```diff
@@ -36,9 +36,9 @@ spec:
...
             requests:
-              memory: 300Mi
+              memory: 600Mi
```
Shwetabhk, line 41: Set requests mem to 300 and limit to 450 max

Shwetabhk (approved):

PR #361, opened 2023-02-23
Placeholder bug solved

ruuushhh:


apps/mappings/tasks.py
```diff
@@ -819,6 +821,9 @@ def create_fyle_expense_custom_field_payload(netsuite_attributes: List[Destinati
             'code': None
         }
 
+        if placeholder_value:
+            expense_custom_field_payload['placeholder']= placeholder_value
```
ashwin1111, line 825: once space before =

ashwin1111 (approved):

PR #371, opened 2023-04-12
5xx error resolved

ruuushhh:


apps/workspaces/tasks.py
```diff
@@ -183,7 +189,7 @@ def run_email_notification(workspace_id):
                     'netsuite_subsidiary': netsuite_subsidiary,
                     'workspace_id': workspace_id,
                     'year': date.today().year,
-                    'export_time': workspace.last_synced_at.date(),
+                    'export_time': export_time.date(),
```
ashwin1111, line 192: let's add the safety check we did in intacct, if it doesn't exist - take current date

ashwin1111 (approved):

PR #372, opened 2023-04-14
Sentry integration

ruuushhh:


.github/workflows/pytest_action.yml
```diff
@@ -16,20 +16,15 @@ jobs:
         run: |
           docker-compose -f docker-compose-pipeline.yml build
           docker-compose -f docker-compose-pipeline.yml up -d
-          docker-compose -f docker-compose-pipeline.yml exec -T api pytest tests/ --cov --junit-xml=test-reports/report.xml --cov-report=term-missing --cov-fail-under=94 | tee pytest-coverage.txt
+          docker-compose -f docker-compose-pipeline.yml exec -T api pytest tests/ --cov --cov-report=xml --cov-fail-under=94
```
ashwin1111, line 19: TODO: check if bringing down coverage fails the action and blocks merge

codecov.yml
```diff
@@ -0,0 +1,17 @@
+codecov:
+  require_ci_to_pass: false
+comment:
+  behavior: default
+  layout: reach, diff, flags, files
+  require_base: false
+  require_changes: false
+  require_head: false
+  show_carryforward_flags: false
```
ashwin1111, line 9: can you add a one liner comment for each configuration, would easier for people to read and understand on a top level without having to reading docs

ashwin1111 (approved):

PR #385, opened 2023-05-15
Sync option in Netsuite Integration page should sync in all the dimensions.

Ashutosh619-sudo:


apps/netsuite/views.py
```diff
@@ -196,6 +198,19 @@ def post(self, request, *args, **kwargs):
             workspace = Workspace.objects.get(pk=kwargs['workspace_id'])
 
             netsuite_credentials = NetSuiteCredentials.objects.get(workspace_id=workspace.id)
+
+            project_mappings = MappingSetting.objects.filter(workspace_id=workspace.id)
```
ashwin1111, line 202: name it mapping_settings, since we are not really filtering by project

apps/netsuite/views.py
```diff
@@ -196,6 +198,19 @@ def post(self, request, *args, **kwargs):
...
+            for project_mapping in project_mappings:
+                if project_mapping.source_field == "PROJECT":
+                    # run auto_import_and_map_fyle_fields
+                    async_task("apps.mappings.tasks.auto_import_and_map_fyle_fields",int(workspace.id))
```
ashwin1111, line 206: one space after comma (,)

apps/netsuite/views.py
```diff
@@ -196,6 +198,19 @@ def post(self, request, *args, **kwargs):
...
+                if project_mapping.source_field == "COST CENTER":
```
ashwin1111, line 207: elif

apps/netsuite/views.py
```diff
@@ -196,6 +198,19 @@ def post(self, request, *args, **kwargs):
...
+                    # run auto_create_cost_center_mappings
+                    async_task("apps.mappings.tasks.auto_create_cost_center_mappings",int(workspace.id))
+                if project_mapping.is_custom:
```
ashwin1111, line 210: elif

apps/netsuite/views.py
```diff
@@ -196,6 +198,19 @@ def post(self, request, *args, **kwargs):
...
+                    # run async_auto_create_custom_field_mappings
+                    async_task("apps.mappings.tasks.async_auto_create_custom_field_mappings",int(workspace.id))
```
ashwin1111, line 212: instead of doing 3 async task, we should do chain. Checkout https://django-q.readthedocs.io/en/latest/chain.html / references in same repo for usage

apps/netsuite/views.py
```diff
@@ -196,6 +198,22 @@ def post(self, request, *args, **kwargs):
...
+                    # run auto_import_and_map_fyle_fields
+                    chain.append('apps.mappings.tasks.auto_import_and_map_fyle_fields', int(workspace.id))
+                elif mapping_setting.source_field == 'COST_CENTER':
+                    # run auto_create_cost_center_mappings
+                    chain.append('apps.mappings.tasks.auto_create_cost_center_mappings', int(workspace.id))
+                elif mapping_setting.is_custom:
+                    # run async_auto_create_custom_field_mappings
+                    chain.append('apps.mappings.tasks.async_auto_create_custom_field_mappings', int(workspace.id))
+            
+            chain.run()
```
ashwin1111, line 216: do this conditionally only if chain has at least 1 task, there can be orgs where there are no mapping settings.

ashwin1111 (approved):

PR #406, opened 2023-07-07
Exempting HTML files from getting exported to NetSuite

Shwetabhk:


apps/netsuite/tasks.py
```diff
@@ -95,9 +95,15 @@ def load_attachments(netsuite_connection: NetSuiteConnector, expense: Expense, e
...
             })
-            files_list = [{'id': file_ids[0]}]
+            
+            for file_id in file_ids:
+                files_list.append({'id': file_id})
+
             attachments = platform.files.bulk_generate_file_urls(files_list)
 
+            # Filter HTML attachments
+            attachments = list(filter(lambda attachment: attachment['content_type'] != 'text/html', attachments))
```
ashwin1111, line 105: we should also limit it to 1st element, in case of multiple pdfs

ashwin1111 (approved):

NileshPant1999 (approved):

ashwin1111 (approved):

PR #412, opened 2023-07-27
Added check for department id

ruuushhh:


NileshPant1999 (approved):

apps/netsuite/models.py
```diff
@@ -581,7 +581,8 @@ def create_credit_card_charge(expense_group: ExpenseGroup):
 
         if general_mappings.use_employee_department and general_mappings.department_level in (
             'ALL', 'TRANSACTION_BODY') and employee_field_mapping == 'EMPLOYEE':
-            department_id = employee_mapping.destination_employee.detail.get('department_id')
+            if employee_mapping.destination_employee.detail.get('department_id'):
```
ashwin1111, line 584: use the same parent if, no point in using 2 if

ashwin1111 (approved):

PR #418, opened 2023-08-28
Adding change to do incremental employee fetch from NetSuite

Shwetabhk:


apps/netsuite/connector.py
```diff
@@ -647,7 +648,22 @@ def sync_employees(self):
         """
         subsidiary_mapping = SubsidiaryMapping.objects.get(workspace_id=self.workspace_id)
 
-        employees_generator = self.connection.employees.get_all_generator()
+        max_updated_at = DestinationAttribute.objects.filter(
+            workspace_id=self.workspace_id).all().aggregate(
```
ashwin1111, line 652: add attribute type filter

apps/netsuite/connector.py
```diff
@@ -647,7 +648,22 @@ def sync_employees(self):
...
+            max_updated_at=Max('updated_at')
+        )['max_updated_at']
+
+        if max_updated_at:
+            search_value = (max_updated_at - timedelta(days=30)).isoformat()
+
+            last_modified_date_query = {
```
ashwin1111, line 659: handle new org case, last_modified_date_query is assigned inside if block

apps/netsuite/connector.py
```diff
@@ -647,7 +648,22 @@ def sync_employees(self):
         """
         subsidiary_mapping = SubsidiaryMapping.objects.get(workspace_id=self.workspace_id)
 
-        employees_generator = self.connection.employees.get_all_generator()
+        max_updated_at = DestinationAttribute.objects.filter(
+            workspace_id=self.workspace_id).all().aggregate(
+            max_updated_at=Max('updated_at')
```
ashwin1111, line 653: will it be updated_at=Max('updated_at') or max_column_name is the syntax?

Shwetabhk: max_column_name is the syntax

PR #433, opened 2023-10-25
V2 Export setting api

Ashutosh619-sudo:


apps/workspaces/apis/export_settings/serializers.py
```diff
@@ -0,0 +1,174 @@
...
+            'default_ccc_account',
+            'accounts_payable',
+            'default_ccc_vendor',
+        ]
+
+    def get_reimbursable_account(self, instance: GeneralMapping):
+        return {
+            'id': instance.reimbursable_account_id,
+            'name': instance.reimbursable_account_name 
+        }
```
NileshPant1999, line 55: can we add linespaces between each function

NileshPant1999 (approved):

apps/workspaces/apis/export_settings/serializers.py
```diff
@@ -0,0 +1,174 @@
...
+class ConfigurationSerializer(serializers.ModelSerializer):
+    
+    class Meta:
+        model = Configuration
+        fields = [
+            'reimbursable_expenses_object',
+            'corporate_credit_card_expenses_object',
+            'is_simplify_report_closure_enabled',
+            'auto_map_employees',
+            'employee_field_mapping'
```
ashwin1111, line 31: this is already a part of map_employees, not needed here

PR #448, opened 2023-11-14
Setting and Resolving Netsuite Error

Ashutosh619-sudo:


apps/netsuite/exceptions.py
```diff
@@ -25,6 +25,18 @@ def __handle_netsuite_connection_error(expense_group: ExpenseGroup, task_log: Ta
         'expense_group_id': expense_group.id,
         'message': 'NetSuite Account not connected'
     }
+
+    Error.objects.update_or_create(
+                    workspace_id=expense_group.workspace_id,
+                    expense_group=expense_group,
```
NileshPant1999, line 31: spacing is not correct

NileshPant1999 (approved):

PR #459, opened 2023-11-23
Sync import API

Ashutosh619-sudo:


apps/fyle/views.py
```diff
@@ -419,3 +419,20 @@ def get(self, request, *args, **kwargs):
                 },
                 status=status.HTTP_400_BAD_REQUEST
             )
+
+
+class ExpenseGroupSyncView(generics.CreateAPIView):
```
NileshPant1999, line 424: where is the the corresponding url path in url file ??

ashwin1111 (approved):

PR #460, opened 2023-11-28
feat: Allow intercompany vendors

Shwetabhk:


apps/netsuite/connector.py
```diff
@@ -545,7 +546,7 @@ def sync_vendors(self):
                 detail = {
                     'email': vendor['email'] if vendor['email'] else None
                 }
-                if 'subsidiary' in vendor and vendor['subsidiary']:
+                if 'subsidiary' in vendor and vendor['subsidiary'] or configuration.allow_intercompany_vendors:
```
ashwin1111, line 549: `if 'subsidiary' in vendor and vendor['subsidiary'] or (configuration and configuration.allow_intercompany_vendors)`

there are chances this func would get executed before config is created

ashwin1111 (approved):

apps/netsuite/connector.py
```diff
@@ -545,7 +546,7 @@ def sync_vendors(self):
...
                     if vendor['subsidiary']['internalId'] == subsidiary_mapping.internal_id:
```
ashwin1111, line 550: this needs fixing

ruuushhh (approved):

PR #462, opened 2023-11-29
Fix: Expense custom Field active flag

ruuushhh:


apps/mappings/tasks.py
```diff
@@ -968,7 +968,7 @@ def upload_attributes_to_fyle(workspace_id: int, netsuite_attribute_type: str, f
         'workspace_id': workspace_id
     }
 
-    if netsuite_attribute_type not in DEFAULT_NETSUITE_IMPORT_TYPES or netsuite_attribute_type == 'DEPARTMENT':
+    if netsuite_attribute_type not in DEFAULT_NETSUITE_IMPORT_TYPES or netsuite_attribute_type == 'DEPARTMENT' or netsuite_attribute_type == 'CLASS':
```
NileshPant1999, line 971: just use netsuite_attribute_type__in

NileshPant1999: also why are we not removing department and class from the list

NileshPant1999 (approved):

PR #465, opened 2023-11-29
Export Url Bug Fix

Ashutosh619-sudo:


fyle_netsuite_api/utils.py
```diff
@@ -36,9 +36,10 @@ def filter_queryset(self, queryset):
         return super().filter_queryset(queryset)
     
 
-def generate_netsuite_export_url(response_logs, ns_account_id):
+def generate_netsuite_export_url(response_logs, netsuite_creds):
```
ashwin1111, line 39: `netsuite_credentials`. Also add type for args. Doc strings too.

ashwin1111 (approved):

PR #476, opened 2024-01-08
resolved tax group rate issue for CA

Ashutosh619-sudo:


apps/netsuite/connector.py
```diff
@@ -902,9 +902,13 @@ def sync_tax_items(self):
         for tax_groups in tax_groups_generator:
             for tax_group in tax_groups:
                 if not tax_group['isInactive'] and tax_group['itemId']:
+                    if tax_group['nexusCountry']['internalId'] == 'CA':
```
ashwin1111, line 905: `if tax_group['nexusCountry'] and tax_group['nexusCountry']['internalId'] == 'CA':` to be safe

apps/netsuite/connector.py
```diff
@@ -902,9 +902,13 @@ def sync_tax_items(self):
...
+                        unit_price1 = float(tax_group['unitprice1'][:-1])
```
ashwin1111, line 906: check for unitprice1 existence else default to 0, same for unit_price2 too

ashwin1111 (approved):

ashwin1111 (approved):

PR #491, opened 2024-02-12
Update export status in Fyle

ruuushhh:


apps/netsuite/tasks.py
```diff
@@ -72,6 +74,36 @@
...
+    :return: None
+    """
+    try:
+        url = expense_group.export_url
+    except Exception as error:
+        # Defaulting it to Netsuite app url, worst case scenario if we're not able to parse it properly
+        netsuite_credentials = NetSuiteCredentials.objects.get(workspace_id=expense_group.workspace_id)
+        ns_account_id = netsuite_credentials.ns_account_id.lower()
+        url = f'https://{ns_account_id}.app.netsuite.com/app/'
+        logger.error('Error while generating export url %s', error)
```
ashwin1111, line 103: there is no chance of exception here, can remove this block

apps/netsuite/tasks.py
```diff
@@ -72,6 +74,36 @@
...
+    :param workspace_id: Workspace ID
+    :param fund_source: Fund source
+    :return: None
+    """
+    fyle_org_id = Workspace.objects.get(pk=workspace_id).fyle_org_id
+    update_expenses_in_progress(in_progress_expenses)
+    post_accounting_export_summary(fyle_org_id, workspace_id, fund_source)
+
+
+def generate_export_url_and_update_expense(expense_group: ExpenseGroup) -> None:
```
ashwin1111, line 90: rename func too

ashwin1111 (approved):

apps/netsuite/tasks.py
ruuushhh: Have removed this function
and updated to 
`update_complete_expenses(expense_group.expenses.all(), expense_group.export_url)`

PR #494, opened 2024-02-12
New Import Q Cluster

ruuushhh:


apps/fyle/tasks.py
```diff
@@ -50,7 +50,7 @@ def schedule_expense_group_creation(workspace_id: int):
     
     task_log, fund_source, configuration = get_task_log_and_fund_source(workspace_id)
 
-    async_task('apps.fyle.tasks.create_expense_groups', workspace_id, configuration, fund_source, task_log)
+    async_task('apps.fyle.tasks.create_expense_groups', workspace_id, configuration, fund_source, task_log, q_options={'cluster': 'import'})
```
Shwetabhk, line 53: This should be in the default queue, if import is stuck this will also block exports

Shwetabhk (approved):

PR #497, opened 2024-02-14
Added RetryException handler and bumped the sdk versions

Hrishabh17:
Please review.


requirements.txt
```diff
@@ -21,7 +21,7 @@ django-sendgrid-v5==1.2.0
 enum34==1.1.10
 future==0.18.2
 fyle==0.35.0
```
ashwin1111, line 23: can you bump this too, to be safe

ashwin1111 (approved):

PR #506, opened 2024-03-12
Resource Project: refactored imports

Hrishabh17:


apps/mappings/constants.py
```diff
@@ -51,3 +51,15 @@
     'VENDOR',
     'EMPLOYEE'
 ]
+
+SYNC_METHODS = {
+    'ACCOUNT': 'accounts',
+    'ITEM': 'items',
+    'TAX_ITEMS': 'tax_items',
```
ashwin1111, line 58: for tax we should sync 2 types

ashwin1111 (approved):

PR #511, opened 2024-03-14
Refactor import category

Hrishabh17:


apps/workspaces/apis/import_settings/triggers.py
```diff
@@ -80,9 +79,6 @@ def post_save_configurations(self, configurations_instance: Configuration):
             mapping_settings=self.__mapping_settings
         )
 
-        if not configurations_instance.import_items:
-            async_task('apps.mappings.tasks.disable_category_for_items_mapping', q_options={'cluster': 'import'})
```
ashwin1111, line 84: we'll discuss about this tomo

apps/workspaces/signals.py
```diff
@@ -46,10 +46,6 @@ def run_post_configration_triggers(sender, instance: Configuration, **kwargs):
 
     schedule_payment_sync(configuration=instance)
 
-    if not instance.import_items:
-        async_task('apps.mappings.tasks.disable_category_for_items_mapping', instance, q_options={'cluster': 'import'})
```
ashwin1111, line 50: same as above

apps/mappings/queue.py
```diff
@@ -61,13 +62,20 @@ def construct_tasks_and_chain_import_fields_to_fyle(workspace_id: int):
...
+        configurations.corporate_credit_card_expenses_object in ('BILL', 'CHARGE_CARD_TRANSACTION', 'JOURNAL_ENTRY'):
+            is_ccc_mapping_enabled = True
 
         task_settings['import_categories'] = {
             'destination_field': destination_field,
             'destination_sync_methods': destination_sync_methods,
             'is_auto_sync_enabled': True,
-            'is_3d_mapping': False,
-            'charts_of_accounts': []
+            'is_3d_mapping': True,
```
ashwin1111, line 75: Can we merge is_3d_mapping and is_ccc_mapping_enabled
I feel 1 is enough to make things work
for table we already have use_mapping_table

ashwin1111 (approved):

PR #532, opened 2024-04-09
Added condition for inactive attribute

Hrishabh17:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved attribute synchronization by ensuring only active fields are considered.
	- Enhanced subsidiary data handling in location and vendor synchronization processes.
	- Refactored `get_all_locations` function to change the structure of returned data.
	- Adjusted code coverage threshold in `.github/workflows/codecov.yml` and `.github/workflows/pytest_action.yml`.


apps/netsuite/connector.py
```diff
@@ -329,19 +329,20 @@ def get_custom_segment_attributes(self, attribute_type: str, internal_id: str):
             }
 
         for field in custom_records:
-            custom_segment_attributes.append(
```
ashwin1111, line 332: let's skip custom segment, this has some auto sync logic written. Active=True would take care of not importing inactive ones.

apps/netsuite/connector.py
```diff
@@ -376,19 +377,20 @@ def get_custom_record_attributes(self, attribute_type: str, internal_id: str):
...
```
ashwin1111, line 379: same as above

ashwin1111 (approved):

PR #540, opened 2024-04-19
Handled the zeep fault exception during import and export

Hrishabh17:
## Summary by CodeRabbit

- **New Features**
	- Improved error handling in the expense reporting feature to provide clearer error messages and update statuses when specific issues occur.
- **Tests**
	- Enhanced testing for the expense reporting feature to cover additional error scenarios.


apps/netsuite/connector.py
```diff
@@ -1692,6 +1693,9 @@ def post_expense_report(
             created_expense_report = self.connection.expense_reports.post(expense_report_payload)
             return created_expense_report
 
+        except zeep_exceptions.Fault as exception:
```
ashwin1111, line 1696: need not handle here, handle it in netsuite/exceptions, this will silence the exception and it might end up in IN_PROGRESS

apps/netsuite/exceptions.py
```diff
@@ -148,6 +149,14 @@ def wrapper(*args):
                 if not payment:
                     update_failed_expenses(expense_group.expenses.all(), False)
 
+            except zeep_exceptions.Fault as exception:
+                logger.info('Error in Expense Report creation: %s', exception.__dict__)
```
ashwin1111, line 153: remove `Expense Report creation` as this is a generic func, can make it error while exporting

ashwin1111 (approved):

PR #551, opened 2024-04-29
Parse existing error

Ashutosh619-sudo:


scripts/python/parse-error.py
```diff
@@ -0,0 +1,31 @@
+from apps.netsuite.errors import error_matcher, get_entity_values, replace_destination_id_with_values
+from apps.tasks.models import Error
+from apps.workspaces.models import Configuration
+
+errors_count = Error.objects.filter(type='NETSUITE_ERROR', is_resolved=False).count()
```
ashwin1111, line 5: remove is_resolved filter, let's run for all. And add is_parsed=False

PR #557, opened 2024-05-02
increase repetition count for every error repeated

NileshPant1999:
## Summary by CodeRabbit

- **New Features**
  - Added a `repetition_count` field to track error occurrences effectively.

- **Enhancements**
  - Improved error handling by automatically incrementing the `repetition_count` for specific tasks and exceptions.

- **Bug Fixes**
  - Corrected imports to ensure the proper functioning of updated features.

- **Refactor**
  - Updated various functions across the application to utilize the new `repetition_count` field efficiently.

- **Documentation**
  - Added helpful text for the new `repetition_count` field to assist in understanding its purpose.


apps/netsuite/exceptions.py
```diff
@@ -51,6 +52,9 @@ def __handle_netsuite_connection_error(expense_group: ExpenseGroup, task_log: Ta
                 'is_resolved': False
             })
 
+        error.repetition_count = error.repetition_count + 1
```
ashwin1111, line 55: we should increment only when the error message is different, else we should reset to 0

ashwin1111, line 55: also, do this only for updates and skip creates. Write 1 smoll func in Error class so that same logic is not repeated everywhere.

ashwin1111 (approved):

PR #587, opened 2024-06-12
Increase timeout to 20min

Hrishabh17:


fyle_netsuite_api/settings.py
```diff
@@ -211,7 +211,7 @@
     'max_attempts': 1,
     'attempt_count': 1,
     'retry': 14400,
-    'timeout': 900, # 15 mins
+    'timeout': 1200, # 20 mins
```
ashwin1111, line 214: this is export worker, need not change this

ashwin1111 (approved):

PR #655, opened 2024-10-11
feat: Single credit line for Journal Entry

ashwin1111:
### Description

## Clickup
https://app.clickup.com/t/86cwrqeuy


ashwin1111:
tests added to next PR, @ruuushhh https://github.com/fylein/fyle-netsuite-api/pull/656

apps/netsuite/connector.py
```diff
@@ -2085,8 +2085,89 @@ def construct_journal_entry_lineitems(self, journal_entry_lineitems: List[Journa
...
+                'customFieldList': [],
+                'debit': None,
+                'debitTax': None,
+                'eliminate': None,
+                'endDate': None,
+                'grossAmt': None,
+                'line': None,
+                'lineTaxCode': None,
+                'lineTaxRate': None,
+                'memo': 'Total Credit Amount',
```
ruuushhh, line 2150: One question should we also add the unique identifier for this 

For example we have 2 credit lines
123-ACB
456-XYZ

If both will have same memo then there is no referral to memo
We should add something like `Total Credit Amount for ACC/Entity`

apps/netsuite/connector.py
```diff
@@ -2085,8 +2085,89 @@ def construct_journal_entry_lineitems(self, journal_entry_lineitems: List[Journa
...
+                'totalAmount': None,
+            }
+
+            lines.append(lineitem)
+
+        return lines
+
     def __construct_journal_entry(self, journal_entry: JournalEntry,
-                                  journal_entry_lineitems: List[JournalEntryLineItem]) -> Dict:
+                                  journal_entry_lineitems: List[JournalEntryLineItem], configuration: Configuration) -> Dict:
```
ruuushhh, line 2170: why not call configuration here only instead of passing it as a param ??

ruuushhh (approved):

apps/netsuite/connector.py
ashwin1111: account and entity info is present in same line, hence didn't add - will check with Hemanth once

apps/netsuite/connector.py
ashwin1111: no extra db call when we have the value already present

ruuushhh: Okay got it

PR #661, opened 2024-11-11
feat: handle posted at

anishfyle:
### Description
- Handle posted at, added flag to check if posted at is null or not, if it is null then we get spent at value in posted at

## Clickup
https://app.clickup.com/t/86cx1az88

## Summary by CodeRabbit

- **New Features**
	- Introduced a new flag, `is_posted_at_null`, in the Expense model to indicate if the `posted_at` field is null, enhancing tracking capabilities.
	- Updated the SQL database schema to include the new `is_posted_at_null` column in the expenses table.
	- Changed the default value of the `change_accounting_period` field in the Configuration model to `True`, indicating that the accounting period will now change by default.
- **Bug Fixes**
	- Modified the test fixture to clear existing TaskLog entries before creating new ones, preventing potential conflicts.


apps/fyle/models.py
```diff
@@ -107,6 +107,7 @@ class Expense(models.Model):
     spent_at = models.DateTimeField(null=True, help_text='Expense spent at')
     approved_at = models.DateTimeField(null=True, help_text='Expense approved at')
     posted_at = models.DateTimeField(null=True, help_text='Date when the money is taken from the bank')
+    is_posted_at_null = models.BooleanField(default=False, help_text='Flag check if posted at is null or not')
```
ruuushhh, line 110: please add test to cover the posted_at <> spent_at change and for this flag as well

ruuushhh (approved):

PR #662, opened 2024-11-11
feat: Netsuite Tax Override

ruuushhh:
### Description
Netsuite Tax Override

## Clickup
https://app.clickup.com/1864988/v/l/6-901604646819-1

## Summary by CodeRabbit

- **New Features**
	- Introduced a new setting to enable tax balancing for general mappings.
	- Enhanced tax handling for line items in the NetSuite integration, allowing for better management of taxed and untaxed entries.
	- Added detailed tax information structures to improve clarity in tax-related data.

- **Bug Fixes**
	- Improved accuracy in tax calculations across various transaction types, ensuring correct processing of bills, credit card charges, and journal entries.
	- Enhanced error handling in tests to ensure proper responses to various failure scenarios.


apps/netsuite/connector.py
```diff
@@ -1220,13 +1280,34 @@ def construct_bill_lineitems(
...
+
+            base_line = {
                 'orderDoc': None,
                 'orderLine': None,
                 'line': None,
-                'amount': line.amount - line.tax_amount if (line.tax_item_id and line.tax_amount is not None) else line.amount,
-                'grossAmt': None if override_tax_details else line.amount,
-                'taxDetailsReference': expense.expense_number if override_tax_details else None,
+                'amount': line.amount,
+                'grossAmt': line.amount,
```
ashwin1111, line 1309: override_tax_details to be handled

apps/netsuite/connector.py
```diff
@@ -1220,13 +1280,34 @@ def construct_bill_lineitems(
...
+                'taxDetailsReference': None,
```
ashwin1111, line 1310: same as above

apps/netsuite/connector.py
```diff
@@ -1254,10 +1335,10 @@ def construct_bill_lineitems(
                 'customFieldList': netsuite_custom_segments,
                 'isBillable': line.billable,
                 'tax1Amt': None,
-                'taxAmount': line.tax_amount if (line.tax_item_id and line.tax_amount is not None and not override_tax_details) else None,
+                'taxAmount': None,
```
ashwin1111, line 1338: test exports with tax amount as none, tax amount should get populated in NS

apps/netsuite/connector.py
```diff
@@ -1161,42 +1162,101 @@ def sync_customers(self):
...
+    
+    def handle_taxed_line_items(self, base_line, line, workspace_id, export_module):
+        """
+        Handle line items where tax is applied or modified by the user.
+        :param base_line: The base line item template that will be modified.
+        :param line: The original line with tax and amount information.
+        :param is_credit_card_charge: Boolean flag to differentiate between credit card charges and other transactions.
+        :return: List of lines (taxed and/or untaxed).
+        """
+        general_mapping: GeneralMapping = GeneralMapping.objects.get(workspace_id=workspace_id)
```
ashwin1111, line 1174: re use this var from parent func

apps/netsuite/connector.py
```diff
@@ -1161,42 +1162,101 @@ def sync_customers(self):
...
+
+            if export_module == 'JOURNAL_ENTRY':
+                taxable_line['grossAmt'] = None
+                taxable_line['debit'] = recalculated_net_amount
+                taxable_line.pop('amount', None)
+                untaxed_line['grossAmt'] = None
+                untaxed_line['debit'] = untaxed_amount
+                untaxed_line.pop('amount', None)
+            
+            if export_module not in ('CREDIT_CARD_CHARGE', 'BILL'):
```
ashwin1111, line 1209: just for readability do, in (ER, JE)

ashwin1111 (approved):

apps/netsuite/connector.py
ruuushhh: I can do this but this will add parameter to a lot of function 
For example if I want to add this to bill

Then 
post_bill (This has configuration) => _construct_bill => construct_bill_line_items => handle_taxed_line_items

I think this is alot to just get one setting, but will differ to you

apps/netsuite/connector.py
ruuushhh: this is already handled below

apps/netsuite/connector.py
ruuushhh: this is already handled below

apps/netsuite/connector.py
ruuushhh: have tested before will test again, No need but if needed will add

apps/netsuite/connector.py
ashwin1111: that is ok, we can save 1 db call per line

PR #717, opened 2025-02-18
feat: skip export v2

anishfyle:
### Description
- skip export v2, using signals

## Clickup
- https://app.clickup.com/t/86cxyw9rw


ashwin1111 (approved):
pls also add today's filtered expense fix @anishfyle

PR #718, opened 2025-02-19
Feat: Add loggers to post vendor

Ashutosh619-sudo:
### Description

## Clickup
Please add link here
app.clickup.com/

## Summary by CodeRabbit

- **Chores**
  - Enhanced monitoring for vendor processing, providing improved tracking of vendor data handling without affecting core functionality.
  - Updated error logging format for clearer output during vendor creation.

### DIFF
--- a/apps/netsuite/connector.py
+++ b/apps/netsuite/connector.py
@@ -755,6 +755,7 @@ class NetSuiteConnector:
 
         vendor_response = None
         try:
+            logger.info('Workspace: %s - Processing vendor payload: %s', self.workspace_id, vendor)
             vendor_response = self.connection.vendors.post(vendor)
         except NetSuiteRequestError as exception:
             logger.info({'error': exception})
--- a/apps/netsuite/connector.py
+++ b/apps/netsuite/connector.py
@@ -755,10 +755,14 @@ class NetSuiteConnector:
 
         vendor_response = None
         try:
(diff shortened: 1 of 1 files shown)


apps/netsuite/connector.py
```diff
@@ -755,6 +755,7 @@ def post_vendor(self, expense_group: ExpenseGroup, vendor: ExpenseAttribute = No
 
         vendor_response = None
         try:
+            logger.info('Workspace: %s - Processing vendor payload: %s', self.workspace_id, vendor)
```
ashwin1111, line 758: log expense_group.id as well and instead of processing call it Vendor POST payload

apps/netsuite/connector.py
```diff
@@ -755,6 +755,7 @@ def post_vendor(self, expense_group: ExpenseGroup, vendor: ExpenseAttribute = No
...
             vendor_response = self.connection.vendors.post(vendor)
         except NetSuiteRequestError as exception:
             logger.info({'error': exception})
```
ashwin1111, line 761: add a note, error in creating vendor and then the exception

ashwin1111:
```
vendor_post_restictions = {'representingsubsidiary': False, 'isPerson': False, 'entityId': False} - default

representingsubsidiary - X
entityId - X

{'representingsubsidiary': True, 'isPerson': False, 'entityId': True}

ashwin1111 (approved):

PR #758, opened 2025-04-16
Feat: Expenses can be skipped after being imported to integration

Ashutosh619-sudo:
### Description

## Clickup
Please add link here
https://app.clickup.com/1864988/v/l/li/901605343641

## Summary by CodeRabbit

- **New Features**
	- Expenses that meet updated skip-export rules are now automatically re-evaluated and marked as skipped when expense filters are changed.
- **Bug Fixes**
	- Improved accuracy in identifying and skipping only relevant expenses, ensuring already skipped expenses are not processed again.
- **Chores**
	- Enhanced error handling and logging for expense filter updates to provide clearer feedback if processing fails.
	- Cleaned up incomplete task logs and errors for expense groups after skip-export rule re-evaluation to maintain data consistency.


apps/fyle/actions.py
```diff
@@ -101,14 +97,17 @@ def mark_expenses_as_skipped(final_query: Q, expenses_object_ids: List, workspac
...
             )
         )
 
-    __bulk_update_expenses(expense_to_be_updated)
-    return [expense.id for expense in expenses_to_be_skipped]
+    if expense_to_be_updated:
+        __bulk_update_expenses(expense_to_be_updated)
+
+    # Return the updated expense objects
+    return expenses_to_be_skipped
```
ashwin1111, line 110: mark_expenses_as_skipped is called elsewhere too, since return type is changed we should fix all callers

apps/fyle/signals.py
```diff
@@ -0,0 +1,28 @@
...
+    :param sender: Sender Class
+    :param instance: Row Instance of Sender Class
+    :return: None
+    """
+    if instance.join_by is None:
+        try:
+            re_run_skip_export_rule(instance.workspace)
+        except Exception as e:
+            logger.error(f'Error while processing expense filter for workspace: {instance.workspace.id} - {str(e)}')
+            raise
```
ashwin1111, line 27: rm this line, we're raising at 28 already

apps/fyle/actions.py
Ashutosh619-sudo: done

apps/fyle/signals.py
Ashutosh619-sudo: done

apps/fyle/tasks.py
```diff
@@ -276,3 +280,76 @@ def update_non_exported_expenses(data: Dict) -> None:
...
+                error = Error.objects.filter(
+                    workspace_id=workspace.id,
+                    expense_group_id=expense_group.id,
+                    type__in=['NETSUITE_ERROR']
+                ).first()
+                if error:
+                    logger.info('Deleting Netsuite error for expense group %s before export', expense_group.id)
+                    error.delete()
+
+                expense_group.expenses.remove(*skipped_expenses)
```
ashwin1111, line 339: does this line not throw error? ex: eg_1 has ex_1, ex_2 and skipped_expenses list has ex_2, ex_3

Ashutosh619-sudo: no it wont Ashwin, its like doing IN (ex_2, ex3), it will ignore which is not there.

ashwin1111 (approved):

PR #762, opened 2025-04-25
feat: Netsuite health check api support

vignesh05904:
### Description
Added health check api support to check weather the netsuite token is expired or healthy. If expired 4xx response, If healthy 2xx response, also an minor database changes applied added new row is_expired to manage token status.

## Clickup
https://app.clickup.com/1864988/v/l/li/901608004149

## Summary by CodeRabbit

* **New Features**
  * Added a health check endpoint to verify the status of NetSuite token connections for each workspace.
  * Introduced automatic invalidation and tracking of expired NetSuite credentials, with a new indicator to mark credentials as expired.

* **Bug Fixes**
  * Improved error handling and messaging for missing or expired NetSuite credentials across multiple features and endpoints.
  * Enhanced system stability by gracefully handling absent or invalid credentials in background processes and API calls.

* **Chores**
  * Updated test suites to reflect new credential handling logic and added comprehensive tests for the new health check and integration settings update.
  * Updated `.gitignore` to exclude `.cursor` files.


apps/workspaces/views.py
```diff
@@ -66,6 +65,37 @@ def get(self, request):
...
+         message = "Netsuite connection expired"
+        else:
+            try:
+                netsuite_connection = NetSuiteConnector(netsuite_credentials=netsuite_credentials, workspace_id=workspace_id)
+                netsuite_connection.connection.locations.count()
+            except Exception as e:
+                status_code = status.HTTP_400_BAD_REQUEST
+                message = "Netsuite connection expired"
+                netsuite_credentials.is_expired = True
+                netsuite_credentials.save()
```
ashwin1111, line 94: let's also have a function similar to qbo - invalidate_qbo_credentials() that sets is expired to false and also post to integrations settings API

ashwin1111: let's also call the new function from handle_netsuite_exceptions() decorator and handle `NetSuiteLoginError` exception. Also whatever function uses @handle_netsuite_exceptions let's make NetSuiteCredential.get() by adding additional is_expired=False filter so that only active token is fetched and expired ones don't proceed

ashwin1111 (commented):
will check PR post fixture is fixed

tests/test_workspaces/test_models.py
```diff
@@ -31,3 +31,55 @@ def test_workspace_creation():
 
     assert workspace.fyle_org_id=='nil123pant'
     assert workspace.name=='Fyle Test Org'
+
+@pytest.mark.django_db(databases=['default'])
+def test_netsuite_credentials_is_expired():
```
ashwin1111, line 36: this test doesn't add any value or increase coverage since we're doing only db calls and not calling any func, let's remove it

tests/test_workspaces/test_tasks.py
```diff
@@ -136,3 +137,68 @@ def test_post_to_integration_settings(mocker):
...
     assert no_exception
+
+
+@pytest.mark.django_db(databases=['default'])
+def test_patch_integration_settings(mocker):
+    """
+    Test patch_integration_settings task
+    """
+
+    workspace = Workspace.objects.create(
```
ashwin1111, line 148: can use existing workspace, no need of creating new workspace - can you check similar apps for this tests?

tests/test_workspaces/test_views.py
```diff
@@ -15,6 +17,69 @@
...
+    mock_connector = mocker.patch('apps.workspaces.views.NetSuiteConnector')
+    mock_instance = MagicMock()
+    mock_connector.return_value = mock_instance
+    mock_instance.connection.locations.count.return_value = 1
+    
+    cache.delete(cache_key)
+    response = api_client.get(url)
+
+    assert response.status_code == 200
+    assert response.data["message"] == "Netsuite connection is active"
```
ashwin1111, line 72: let's also assert for cache and see if cache is present and true

tests/test_workspaces/test_models.py
vignesh05904: oh ok, I though this will increase coverage

ashwin1111 (approved):

